Issue #8, opened 2018-08-06
`rio --version` produces bad output


sheng-liang:
```
sheng@ubuntu-sheng:~/Downloads$ rio --version
rio github.com/containerd/containerd v0.0.2
```

StrongMonkey:
available to test on v0.0.4-rc1.

sheng-liang:
Fix verified.

Issue #47, opened 2018-09-12
endpoint not connecting on service created with port


tfiduccia:
Version - v0.0.3-rc6
Steps:
1.  rio run -p 80/http --name test/svc --scale=3 ibuildthecloud/demo:v1
2.  rio ps
3.  Grab the endpoint once the state of service becomes active
4.  curl -v <endpoint>

results:
curl: (7) Failed to connect to <endpoint>port 80: Connection refused

tfiduccia:
Version v0.0.3
Verified fixed

PR #51, opened 2018-09-17
Fix routing and version print

StrongMonkey:
https://github.com/rancher/rio/issues/8


StrongMonkey:
https://github.com/rancher/rio/issues/47

PR #55, opened 2018-09-19
Vagrant Support

LLParse:
Ready for review


vagrant.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+# resources allocated to each node
+resources:
+  cpu: "1"
+  # a minimum of 1280 MiB is strongly recommended
+  # increase for memory-hungry workloads
+  memory: "2048"
+
+# Rio version to use such as 'v0.0.3' or 'dev'
+version: dev
```
LLParse: We'll want to change this to a release version (or even an RC) as soon as we can.

PR #60, opened 2018-09-28
make volume works and other fixes

StrongMonkey:


StrongMonkey:
have vendor hacks so don't merge.

Issue #53, opened 2018-09-18
rio run --cidfile should be changed or dropped


tfiduccia:
Steps:
1.  rio run -h

Results:  Under options:    
--cidfile value               Write the container ID to the file

Comment: This doesn't make sense in this current context so we need to either drop it or change it.

StrongMonkey:
available to test on v0.0.4-rc1.

tfiduccia:
Version - v0.0.4 rc5
Verified fixed

PR #66, opened 2018-10-04
Volume works

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/53

cli/pkg/up/up.go
```diff
@@ -13,23 +17,43 @@ import (
...
 )
 
-func readFileInPath(cwd, file string) ([]byte, error) {
+func readFileInPath(cwd, file, filePath string) ([]byte, error) {
+	if strings.HasPrefix(filePath, "http") {
+		base, err := url.Parse(filePath)
+		if err != nil {
+			return nil, err
+		}
+		ref, err := url.Parse(filepath.Join(file))
```
ibuildthecloud: when dealing with URLs use path.Join

ibuildthecloud: wait, you are joining one element....

cli/pkg/up/up.go
```diff
@@ -13,23 +17,43 @@ import (
 	template2 "github.com/rancher/rio/pkg/template"
 )
 
-func readFileInPath(cwd, file string) ([]byte, error) {
+func readFileInPath(cwd, file, filePath string) ([]byte, error) {
```
ibuildthecloud: Seems like `cwd` should be either a http:// url or local directory.  I don't think you need a third argument.

cli/pkg/up/up.go
```diff
@@ -13,23 +17,43 @@ import (
...
 	f, err := symlink.FollowSymlinkInScope(file, cwd)
 	if err != nil {
 		return nil, err
 	}
 	return ioutil.ReadFile(f)
 }
 
-func readFiles(files []string, template *template2.Template, promptReplaceFile bool) error {
+func readFiles(filePath string, files []string, template *template2.Template, promptReplaceFile bool) error {
 	cwd, err := os.Getwd()
```
ibuildthecloud: remove this getwd code and have the caller pass in the "relativeTo" which is either `http` url or a assumed to be a local directory

stacks/istio-stack.yml
```diff
@@ -138,3 +138,59 @@ services:
         - INSTANCE_IP=$(self/ip)
         - ISTIO_META_POD_NAME=$(self/name)
         secrets: identity:/etc/certs
+
```
ibuildthecloud: Why did this change?

PR #69, opened 2018-10-05
nfs and volume tests

StrongMonkey:


pkg/deploy/stack/populate/podcontrollers/statefulset.go
```diff
@@ -30,6 +32,9 @@ func statefulSet(stack *input.Stack, service *v1beta1.Service, cp *controllerPar
 			Template: cp.PodTemplateSpec,
 		},
 	}
+	statefulSet.Labels[useTemplatesLabel] = "true"
```
ibuildthecloud: What is this label used for?  You have to be very careful with labels you add to the selector because they can't be modified after the fact.  So I try to just use as few as possible in the selector

stacks/istio-stack.yml
```diff
@@ -137,4 +137,4 @@ services:
         - POD_NAMESPACE=$(self/namespace)
         - INSTANCE_IP=$(self/ip)
         - ISTIO_META_POD_NAME=$(self/name)
-        secrets: identity:/etc/certs
+        secrets: identity:/etc/certs
```
ibuildthecloud: is there actually a difference here?  did you just remove the new line?

controllers/backend/volume/volume.go
```diff
@@ -55,3 +57,43 @@ func (v *volumeController) sync(key string, pvc *v1.PersistentVolumeClaim) error
 	_, err = v.volumes.Update(newVol)
 	return err
 }
+
+func (v *volumeController) Create(pvc *v1.PersistentVolumeClaim) (*v1.PersistentVolumeClaim, error) {
+	if pvc == nil {
+		return pvc, nil
+	}
+
+	if pvc.Labels["rio.cattle.io/use-templates"] == "" {
```
ibuildthecloud: can this be "rio.catle.io/volumetemplate"?  We only put that label on PVC created from VolumeClaimTemplates

controllers/backend/volume/volume.go
```diff
@@ -55,3 +57,43 @@ func (v *volumeController) sync(key string, pvc *v1.PersistentVolumeClaim) error
...
+		return pvc, nil
+	}
+
+	ns := pvc.Namespace
+	serviceName := pvc.Labels["app"]
+	templateName := strings.SplitN(pvc.Name, "-"+serviceName+"-", 2)
+	volTemplate, err := v.volumeLister.Get(ns, templateName[0])
```
ibuildthecloud: for VolumeClaimTemplates we set the following labels 
```golang
		labels := map[string]string{
			"rio.cattle.io/stack":          stack.Stack.Name,
			"rio.cattle.io/workspace":      stack.Stack.Namespace,
			"rio.cattle.io/volumetemplate": volumeTemplate.Name,
		}
```
I think from that you can find the volume template.

controllers/backend/volume/volume.go
```diff
@@ -55,3 +57,43 @@ func (v *volumeController) sync(key string, pvc *v1.PersistentVolumeClaim) error
...
+		return pvc, err
+	}
+
+	volume := &v1beta1.Volume{}
+	volume.Spec = volTemplate.Spec
+	volume.Name = pvc.Name
+	volume.Namespace = pvc.Namespace
+	volume.Spec.Template = false
+	volume.Labels = make(map[string]string)
+	volume.Labels["rio.cattle.io/volume-template"] = volTemplate.Name
```
ibuildthecloud: we have both "rio.cattle.io/volume-template" and "rio.cattle.io/volumetemplate" now.  Can we just change everything to be "rio.cattle.io/volume-template"  (the hyphen is more consistent with other labels)

pkg/deploy/stack/populate/podcontrollers/statefulset.go
ibuildthecloud: I'm thinking this can just be removed if you use the `rio.catle.io/volume-template` label in the conroller.

controllers/backend/volume/volume.go
StrongMonkey: I remember that the label which was set to the volumeTemplateSpec doesn't get populated into the actual PVC though...

controllers/backend/volume/volume.go
```diff
@@ -17,6 +18,7 @@ func Register(ctx context.Context, rContext *types.Context) {
 	}
 
 	rContext.Core.PersistentVolumeClaims("").AddHandler("volume-controller", vc.sync)
+	rContext.Core.PersistentVolumeClaims("").AddLifecycle("volume-template-controller", vc)
```
ibuildthecloud: Move to sync

Issue #54, opened 2018-09-18
Change info for --network in rio run -h


tfiduccia:
Steps:
1.  rio run -h

Results:  There is a option for --network.  Only two options are available for this, default & home.  We need to change the help and validate that in CLI

tfiduccia:
Version v0.0.4 rc1
Verified fixed

Issue #49, opened 2018-09-13
In -h results (default: []) is showing up when it shouldn't


tfiduccia:
Steps:
1.  rio run -h or rio create -h

Results:
(default: []) is showing up in the explanation for a lot of the fields and this is not useful info.  Shouldn't really be there.

vincent99:
Some are `map[]` too.

StrongMonkey:
available to test on v0.0.4-rc1.

tfiduccia:
Version - v0.0.4 rc1
verified fixed

Issue #48, opened 2018-09-13
memory hard limit is giving me memory reservation error


tfiduccia:
steps:
1.  rio run -n tstack2/tservice1 --memory-limit 50[g]

results:  FATA[0000] failed to parse --memory-reservation: invalid size: '50[g]'.  I should be getting a --memory-limit error not a --memory-reservation error.

StrongMonkey:
@tfiduccia the correct format should be `--memory-limit 50g`(without brackets)

tfiduccia:
@StrongMonkey  - yes I know.  This is not about it not working.  It's about the error being wrong.

StrongMonkey:
available to test on v0.0.4-rc1.

tfiduccia:
Version - v0.0.4 rc1
Verified fixed

PR #70, opened 2018-10-05
Fix cli and add up for remote files

StrongMonkey:


StrongMonkey:
address #53 #48 #54 #49

PR #86, opened 2018-10-16
re-enqueue services if endpoint is not ready

StrongMonkey:


controllers/api/domain/domain.go
```diff
@@ -6,6 +6,8 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+
```
ibuildthecloud: WHITE SPACE

controllers/api/domain/domain.go
```diff
@@ -114,7 +118,10 @@ func (g *Controller) sync(key string, svc *v1.Service) error {
 
 	if len(ips) == 0 {
 		ep, err := g.endpointsLister.Get(svc.Namespace, svc.Name)
-		if err != nil {
+		if err != nil && errors.IsNotFound(err) {
+			g.serviceController.Enqueue(svc.Namespace, svc.Name)
```
ibuildthecloud: are you totally sure you won't get in an infinite loop?  Is it possible that we might just want to mask this error and just return nil on NotFound

controllers/api/domain/domain.go
```diff
@@ -51,11 +51,12 @@ func Register(ctx context.Context, rContext *types.Context) {
...
-		rdnsClient:      rdnsClient,
-		endpointsLister: rContext.Core.Endpoints(settings.IstioExternalLBNamespace).Controller().Lister(),
-		nodeLister:      rContext.Core.Nodes("").Controller().Lister(),
-		stackController: rContext.Rio.Stacks("").Controller(),
+		ctx:               ctx,
+		rdnsClient:        rdnsClient,
+		endpointsLister:   rContext.Core.Endpoints(settings.IstioExternalLBNamespace).Controller().Lister(),
+		nodeLister:        rContext.Core.Nodes("").Controller().Lister(),
+		stackController:   rContext.Rio.Stacks("").Controller(),
+		serviceController: rContext.Core.Services(settings.IstioExternalLBNamespace).Controller(),
```
ibuildthecloud: I don't think you need to add serviceController anymore

Issue #63, opened 2018-10-02
Run --user and --group-add aren't working


tfiduccia:
Steps for --user:
1.  rio run -n tsk1/tuser --user demo-user nginx
2. rio kubectl get -n <namespace> -o=json deploy/tuser

Results:
No user is set in k8s

Steps for --group-add
1.  rio run -n tsk1/tgradd --group-add demo-group nginx
2.  rio inspect tsk1/tgradd

Results:
No group is set on rio side and is not being passed to k8s

StrongMonkey:
@tfiduccia the syntax has changed a lot. I did a bit research and k8s takes UID and GID instead of name to poplate its securityContext. So if you do `rio run -n tsk1/tuser --user 10 nginx` and `rio exec tsk1/tuser id` that should give you the current uid. There is no mapping for grouping right now. Going to talk to darren about it.

StrongMonkey:
We are going to change the mapping so that `--user` and `--group` only takes integer and they are mapped to `UID` AND `GID`

tfiduccia:
Version - v0.1.1-rc1
Mapping only takes integers

PR #87, opened 2018-10-16
update types and generated changes

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/63

PR #92, opened 2018-10-19
Let's encrypt integration and publicdomain

StrongMonkey:


StrongMonkey:
Each cluster will get a wildcards certificate `*.xxxxx.lb.rancher.cloud` from let's encrypt.
Todo: implement renew logic for certificates.

PR #94, opened 2018-10-23
add publicdomain and update types

StrongMonkey:


StrongMonkey:
todo: routeset logic was implemented in another pr https://github.com/rancher/rio/pull/76. Have to modify virtualservice code generated from routeset to add public domain.

PR #104, opened 2018-11-05
don't compare hash if cacert is empty

StrongMonkey:
getCacert will return empty if cacerts is trusted by system certs. Don't
compare the hash if cacerts is trusted(in case it is proxyed through
ngrok)

### DIFF
--- a/pkg/clientaccess/clientaccess.go
+++ b/pkg/clientaccess/clientaccess.go
@@ -122,8 +122,10 @@ func ParseAndValidateToken(server, token string) (*Info, error) {
 		return nil, err
 	}
 
-	if ok, hash, newHash := validateCACerts(cacerts, parsedToken.caHash); !ok {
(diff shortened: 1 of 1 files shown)


StrongMonkey:
https://github.com/rancher/rancher/pull/16448 also need this to get ngrok working

pkg/clientaccess/clientaccess.go
```diff
@@ -122,8 +122,10 @@ func ParseAndValidateToken(server, token string) (*Info, error) {
 		return nil, err
 	}
 
-	if ok, hash, newHash := validateCACerts(cacerts, parsedToken.caHash); !ok {
-		return nil, fmt.Errorf("RIO_TOKEN does not match the server %s != %s", hash, newHash)
+	if cacerts != nil {
```
ibuildthecloud: can this change to `len(cacerts) > 0`?

PR #105, opened 2018-11-05
use 8.8.8.8 as nameservers instead of copy from host

StrongMonkey:


image/init
```diff
@@ -112,12 +112,9 @@ ff02::1		ip6-allnodes
...
 EOF
 
-ln -sf /host/etc/resolv.conf /etc/resolv.conf
-if ! grep -q nameserver /host/etc/resolv.conf; then
-    echo Failing back to 8.8.8.8 DNS 1>&2
-    rm -f /etc/resolv.conf
-    echo 'nameserver 8.8.8.8' > /etc/resolv.conf
-fi
+cat > /etc/resolv.conf << EOF
+nameserver 8.8.8.8
```
ibuildthecloud: Can you change to 1.1.1.1.  Do you have any idea if 1.1.1.1 is blocked in china?

StrongMonkey: no idea... but let's go with 1.1.1.1

PR #106, opened 2018-11-06
always set bridge-nf-call-iptables to 1

StrongMonkey:
https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#network-plugin-requirements

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -12,7 +12,9 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


agent/main.go
```diff
@@ -280,3 +288,7 @@ func getConfig(localURL *url.URL) (*AgentConfig, error) {
 		Certificate: tlsCert,
 	}, err
 }
+
+func setSysctl() error {
+	return ioutil.WriteFile(path.Join(base, sysctlBridgeCallIPTables), []byte(strconv.Itoa(1)), 0640)
```
ibuildthecloud: Can you errors.Wrap this to set a message indicating you failed to write 1 to /proc.....  Otherwise the user gets a random "I/O error" with no idea whats up.

scripts/test
```diff
@@ -29,6 +29,7 @@ ENTER_ROOT=$(pwd)/image/root ./bin/rio --debug agent -s https://localhost:7443 -
 
 export PATH=$(pwd)/bin:$PATH
 
+tail -f /var/lib/rancher/rio/agent/agent.log
```
ibuildthecloud: Did you want to background this? (`&`).  This will block the test script for ever.

PR #111, opened 2018-11-13
let's encrypt and public domain

StrongMonkey:


stacks/cert-manager-stack.yml
```diff
@@ -0,0 +1,21 @@
...
+  cert-manager:
+    global_permissions:
+    - '* certmanager.k8s.io/*'
+    - '* extensions/ingresses'
+    - '* configmaps'
+    - '* secrets'
+    - '* events'
+    - '* services'
+    - '* pods'
+    image: "daishan1992/cert-manager:latest"
```
ibuildthecloud: Can you add a setting for this and tokenize in the stack file `${IMAGE}` and pass it in as an answer?

types/apis/certmanager.k8s.io/v1alpha1/types.go
```diff
@@ -0,0 +1,24 @@
+package v1alpha1
+
+import (
+	"github.com/jetstack/cert-manager/pkg/apis/certmanager/v1alpha1"
+	"github.com/rancher/norman/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type Certificate struct {
```
ibuildthecloud: Can we not just import this type?

api/service/store.go
```diff
@@ -35,23 +37,41 @@ func addEndpoint(apiContext *types.APIContext, schema *types.Schema, data map[st
 		return data, nil
 	}
 
+	spaceID, _ := data["spaceId"].(string)
+	space := ""
+	parts := strings.SplitN(spaceID, "-", 2)
```
ibuildthecloud: This will fail if spacename has a `-` in it.  `foo-bar-23452`

api/service/store.go
```diff
@@ -35,23 +37,41 @@ func addEndpoint(apiContext *types.APIContext, schema *types.Schema, data map[st
...
 		expose = true
+		protocol = proto
 		break
 	}
 
 	if expose {
 		data["endpoints"] = []interface{}{
 			map[string]interface{}{
-				"url": fmt.Sprintf("http://%s.%s.%s", name, stackName, domain),
+				"url": fmt.Sprintf("%s://%s.%s", protocol, HashIfNeed(name, stackName, space), domain),
```
ibuildthecloud: Did you make the changes such that I can set a setting for the public ports to be not 80, 443, but something like 8080 and 8443

cli/cmd/create/ports.go
```diff
@@ -85,3 +85,11 @@ func parsePortBinding(spec string) (client.PortBinding, error) {
 
 	return result, nil
 }
+
+func ParsePort(portDef string) (string, string) {
+	split := strings.SplitN(portDef, "/", 2)
```
ibuildthecloud: There is a util method for this `k, v := kv.Split(portDef, "/")`

cli/cmd/publicdomain/add.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	domain, err = spaceClient.PublicDomain.Create(domain)
+	if err != nil {
+		return err
+	}
+	fmt.Println(domain.Name)
+	return nil
+}
+
+func resolveSpaceStackName(name, defaultWorkspace string) (string, string, string) {
```
ibuildthecloud: Is this the same as rio/cli/pkg/stack/ResolveSpaceStackForName?

cli/cmd/publicdomain/add.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/rancher/rio/cli/pkg/clicontext"
+	"github.com/rancher/rio/types/client/space/v1beta1"
+)
+
+type Add struct {
+}
+
+func (a *Add) Run(ctx *clicontext.CLIContext) error {
+	if len(ctx.CLI.Args()) != 3 {
+		return errors.New("Incorrect Usage. Example: `rio domain add NAME DOMAIN_NAME TARGET_SVC`")
```
ibuildthecloud: NAME seems redundant.  Can we just auto generate name and not show it or require it to be set?

cli/cmd/publicdomain/ls.go
```diff
@@ -0,0 +1,86 @@
...
+	spaceClient, err := cluster.Client()
+	if err != nil {
+		return err
+	}
+	publicDomains, err := spaceClient.PublicDomain.List(&types.ListOpts{})
+	if err != nil {
+		return err
+	}
+	writer := table.NewWriter([][]string{
+		{"NAME", "{{.Name}}"},
```
ibuildthecloud: drop name

cli/cmd/publicdomain/ls.go
```diff
@@ -0,0 +1,86 @@
...
+	return writer.Err()
+}
+
+type wrapper struct {
+	clusterClient *clientcfg.Cluster
+	spaceClient   client.SpaceOperations
+}
+
+func (w wrapper) FormatTarget(obj interface{}) (string, error) {
+	if v, ok := obj.(*client.PublicDomain); ok {
```
ibuildthecloud: early exit instead of nesting this whole method.  So
```golang
v, ok := obj.(*client.PublicDomain)
if !ok {
  return "", nil
}
```

cli/cmd/publicdomain/ls.go
```diff
@@ -0,0 +1,86 @@
...
+		w.clusterClient.DefaultWorkspaceName = v.TargetWorkspaceName
+		workspace, err := w.clusterClient.Workspace()
+		if err != nil {
+			return "", nil
+		}
+		workspaceClient, err := workspace.Client()
+		if err != nil {
+			return "", nil
+		}
+		spaces, err := w.spaceClient.List(&types.ListOpts{})
```
ibuildthecloud: This will do N calls.  So if I have 5 domains this will call List 5 times.  you can instead list all the stacks once up front and then after just search through the list.

cli/cmd/publicdomain/rm.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/rancher/rio/cli/cmd/rm"
+	"github.com/rancher/rio/cli/pkg/clicontext"
+	"github.com/rancher/rio/types/client/space/v1beta1"
+)
+
+type Rm struct {
+}
+
+func (r *Rm) Run(ctx *clicontext.CLIContext) error {
+	return rm.Remove(ctx, client.PublicDomainType)
```
ibuildthecloud: Does remove with with the `domain name`, not just the resource Name?

types/apis/space.cattle.io/v1beta1/publicdomain.go
```diff
@@ -0,0 +1,19 @@
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type PublicDomain struct {
```
ibuildthecloud: can we make PublicDomain namespaced and store them all in RioSystemNamespace?  I'm trying to not ever make cluster level objects anymore

controllers/backend/publicdomain/publicdomain.go
```diff
@@ -0,0 +1,120 @@
...
+	return domain, err
+}
+
+func (d *domainController) getNamespace(domain *spacev1beta1.PublicDomain) (string, error) {
+	namespaces, err := d.namespacesLister.List("", labels.Everything())
+	if err != nil {
+		return "", err
+	}
+	for _, ns := range namespaces {
+		if ns != nil && ns.Labels["field.cattle.io/displayName"] == domain.Spec.TargetWorkspaceName {
```
ibuildthecloud: You can just add this check to the List call before as a selector.  Then you will just get the object you want directly

controllers/backend/publicdomain/publicdomain.go
```diff
@@ -0,0 +1,120 @@
...
+			stack, err := d.stackLister.Get(ns.Name, domain.Spec.TargetStackName)
+			if err != nil {
+				return "", err
+			}
+			namespace := namespace.StackToNamespace(stack)
```
ibuildthecloud: You can do `namespace.StackNamespace(ns.Name, domain.Spec.TargetStackName)` and skip getting `stack` from the lister.

api/service/store.go
StrongMonkey: yes, but I forgot to change here to respect the value.

controllers/backend/publicdomain/publicdomain.go
```diff
@@ -0,0 +1,120 @@
...
+	if err := apply.Apply([]runtime.Object{certs.AcmeIssuer()}, nil, "", "acme-cluster-issuer"); err != nil {
+		return domain, err
+	}
+	ns, err := d.getNamespace(domain)
+	if err != nil {
+		return domain, err
+	}
+	// certificate
+	cert := certs.CertificateHttp(domain)
+	if err := apply.Apply([]runtime.Object{cert}, nil, settings.RioSystemNamespace, "certificate"); err != nil {
```
ibuildthecloud: I'm thinking you want a different groupID.  The groupID determines the set of object to watch for modification and prune.  In general the groupID is a unique for the input data.  What will happen with how this is at the moment is that domain foo.com will create `cert` and then bar.com will create `newCert` and prune (delete) `cert`

cli/cmd/publicdomain/add.go
StrongMonkey: not exactly same. I copied the method and changed it so that it can resolve to workspace/stack/name.

controllers/backend/publicdomain/publicdomain.go
```diff
@@ -0,0 +1,120 @@
...
+		return domain, err
+	}
+	service, err := d.serviceLister.Get(ns, domain.Spec.TargetName)
+	if err != nil && !errors.IsNotFound(err) {
+		return domain, err
+	}
+	if err == nil {
+		service.Annotations[istio.PublicDomainAnnotation] = domain.Spec.DomainName
```
ibuildthecloud: This behavior will prevent registering two public domains to the same service/routeset

cli/cmd/publicdomain/rm.go
StrongMonkey: it removes resourceName. If we want to use generated name, we have to look up for domainName.

controllers/backend/publicdomain/publicdomain.go
```diff
@@ -0,0 +1,120 @@
...
+		_, err = d.services.Services(ns).Update(service)
```
ibuildthecloud: Why doesn't this get stuck in an update loop?  Seems like each update you unconditionally update the object which should cause a loop.

controllers/backend/secrets/secrets.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	updateCopy.Data = data
+
+	if _, err := s.secrets.Update(updateCopy); err != nil {
+		return err
+	}
+	logrus.Infof("Rancher wildcards certs %s for domain %s should be ready to use", updateCopy.Name, fmt.Sprintf("*.%s", settings.ClusterDomain.Get()))
+	go func() {
+		// sleep 5 seconds and wait for secrets to be refreshed in gateway pod
+		time.Sleep(time.Second * 5)
```
ibuildthecloud: hmmm.... no better way to do this?

controllers/backend/stack/controller.go
```diff
@@ -131,10 +135,19 @@ func (s *stackDeployController) deploy(obj *v1beta1.Stack) (*v1beta1.Stack, erro
...
 		obj,
 		configs,
 		services,
 		volumes,
 		routes)
 	return obj, err
 }
+
+func getSpace(stack *v1beta1.Stack) string {
+	parts := strings.SplitN(stack.Namespace, "-", 2)
```
ibuildthecloud: `this-has-parsing-issues-with-too-many-dashes`

pkg/data/data.go
```diff
@@ -20,7 +20,10 @@ func AddData(rContext *types.Context, inCluster bool) error {
 		return err
 	}
 
-	return apply.Apply(systemStacks(inCluster), nil, settings.RioSystemNamespace, "rio-system-stacks")
+	if err := apply.Apply(systemStacks(inCluster), nil, settings.RioSystemNamespace, "rio-system-stacks"); err != nil {
```
ibuildthecloud: no actual change here?

StrongMonkey: i am sure previously there was :)....

types/apis/certmanager.k8s.io/v1alpha1/types.go
StrongMonkey: there are some errors from controller complaining about not finding the type so instead i created those crd first

api/service/store.go
```diff
@@ -35,23 +37,41 @@ func addEndpoint(apiContext *types.APIContext, schema *types.Schema, data map[st
...
+	if len(parts) == 2 {
+		space = parts[1]
+	}
+
 	expose := false
+	protocol := ""
 	for _, ports := range convert.ToMapSlice(data[client.ServiceFieldPortBindings]) {
 		proto, _ := ports[client.PortBindingFieldProtocol].(string)
-		if proto != "http" {
+		if proto != "http" && proto != "https" {
```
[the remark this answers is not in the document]
StrongMonkey: then how can we identify which endpoint we showed to user? cause we have to change endpoint's scheme based on protocol.

cli/cmd/publicdomain/add.go
StrongMonkey: :+1:

api/service/store.go
StrongMonkey: How.....shouldn't spaceId always follow `s-randomstring` pattern cause you generate those name?

StrongMonkey:
@ibuildthecloud ready to review. most of the comments have been addressed.

pkg/certs/certs.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/rancher/rio/pkg/apply"
+	"github.com/rancher/rio/pkg/deploy/istio/output"
+	"github.com/rancher/rio/pkg/settings"
+	spacev1beta1 "github.com/rancher/rio/types/apis/space.cattle.io/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	TlsSecretName        = "rancher-certs"
```
ibuildthecloud: change to rio-certs and rio-wildcard.

pkg/certs/certs.go
```diff
@@ -0,0 +1,161 @@
...
+	RancherWildcardCerts = "rancher-wildcard"
```
ibuildthecloud: RioWildcardCerts

PR #132, opened 2018-12-20
Refactor Apply to ObjectSet

StrongMonkey:


controllers/feature/controller.go
```diff
@@ -175,3 +172,18 @@ func (f *featureHandler) isEnabled(name string) bool {
 	_, ok := f.featureState[name]
 	return ok
 }
+
+func (f *featureHandler) getDepFeature(depName, namespace string) (*v1.Feature, error) {
+	start := time.Millisecond * 250
```
ibuildthecloud: why is this needed?

cli/cmd/feature/feature.go
```diff
@@ -3,6 +3,10 @@ package feature
 import (
 	"fmt"
 
+	"github.com/rancher/rio/cli/cmd/up"
+
```
ibuildthecloud: whitespace

controllers/controllers.go
```diff
@@ -3,6 +3,8 @@ package controllers
 import (
 	"context"
 
+	"github.com/rancher/rio/controllers/setting"
+
```
ibuildthecloud: whitespace

controllers/setting/setting.go
```diff
@@ -0,0 +1,112 @@
+package setting
+
+import (
+	"context"
+	"os"
+	"strings"
+
+	"github.com/rancher/rio/pkg/settings"
+
```
ibuildthecloud: whitespace

features/routing/controllers/routeset/controller.go
```diff
@@ -3,6 +3,8 @@ package routeset
 import (
 	"context"
 
+	"github.com/rancher/rio/features/routing/controllers/util"
+
```
ibuildthecloud: whitespace

features/routing/controllers/service/controller.go
```diff
@@ -3,6 +3,8 @@ package service
 import (
 	"context"
 
+	"github.com/rancher/rio/features/routing/controllers/util"
+
```
ibuildthecloud: whitespace

pkg/features/controller.go
```diff
@@ -120,3 +121,7 @@ func (f *FeatureController) Start(ctx context.Context, feature *v1.Feature) erro
 
 	return nil
 }
+
+func (f *FeatureController) Priority() int {
+	return f.PriorityWeight
```
ibuildthecloud: Why do we need a priority?

types/apis/project.rio.cattle.io/v1/zz_generated_deepcopy.go
```diff
@@ -258,3 +258,61 @@ func (in *PublicDomainSpec) DeepCopy() *PublicDomainSpec {
 	in.DeepCopyInto(out)
 	return out
 }
+
```
ibuildthecloud: can you seperate generated code to a different commit

controllers/feature/controller.go
StrongMonkey: This is just to avoid `not found` error. The first try will always return `not found` error because the race condition when feature is not actually created.

pkg/features/controller.go
StrongMonkey: want to have better control over the order of features

PR #148, opened 2019-01-14
Fix random seeding, definitely maybe

tfiduccia:


tests/conftest.py
```diff
@@ -1,27 +1,30 @@
...
+
+random.seed(os.urandom(8))
 
 
 @pytest.fixture(scope="module")
 def stack():
-    name = "tstk" + str(randint(1000, 5000))
-    system("rio stack create %s" % name)
-    system("rio wait %s" % name)
+    name = "tstk" + str(random.randint(1000, 5000))
```
ibuildthecloud: I would just increase this to 100000,999999

PR #149, opened 2019-01-15
fix bug when turning on and off features

StrongMonkey:


pkg/stackobject/controller.go
```diff
@@ -46,7 +51,9 @@ func NewGeneratingController(ctx context.Context, rContext *types.Context, name
 		stacksCache:    rContext.Rio.Stack.Cache(),
 		namespaceCache: rContext.Core.Namespace.Cache(),
 		injector:       injector,
+		clientAccessor: client,
```
ibuildthecloud: just save the indexer object, not the clientAccesor

PR #243, opened 2019-05-22
Import the auth plugin package for gcp

bstick12:
The auth plugin package for GCP isn't included not allowing Rio to be installed to GKE clusters. #234 

Vendor packaging needs to be updated - unsure if this should be done as part of commit.


StrongMonkey:
@bstick12 Thanks for your contribution! We use trash[https://github.com/rancher/trash](url) to manage go dependency internally. So if you download and run `trash` at the root directory of Rio project it should update the dependency for you.

PR #265, opened 2019-05-23
Rewrite some if-else chains to switch statements

sosiska:
This commit rewrite some if-else chains to switch statements.
Based on Go style guide:
https://golang.org/doc/effective_go.html#switch


sosiska:
Also, there is a question about [this](https://github.com/rancher/rio/blob/master/pkg/apis/common/state.go#L121).

sosiska:
Hello?

PR #286, opened 2019-06-04
Honor multipath KUBECONFIG environment

deas:
Multipath `KUBECONFIG` environment fails:

```
$ rio ps
no config found
...
```
This can be pretty confusing.

This PR aims at a better configuration handling. Personally, I think `c.findKubeConfig()` is still a a bit confusing and we should probably stay as close to `kubectl` behavior as possible.

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -115,7 +115,6 @@ func main() {
 		cli.StringFlag{
 			Name:        "kubeconfig",
 			Usage:       "Kubeconfig file to use",
-			EnvVar:      "KUBECONFIG",
 			Destination: &cfg.Kubeconfig,
 		},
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
@deas The intention is that rio should treat kubeconfig lookup exactly the same as kubectl, so let me look at findKubeConfig() and see where it differs

ibuildthecloud:
Oh wow, I didn't know you "multipath" config was a thing.  I get it now.  Will fix, thanks!

ibuildthecloud:
@deas I rewrote the findKubeConfig to be more inline with kubectl behavior in https://github.com/rancher/rio/pull/293

Issue #163, opened 2019-02-13
--user not being populated in k8s container/securityContext section


tfiduccia:
Steps:
1.  rio run -n stack/service --user demo-user nginx
2.  rio inspect stack/service (notice that user is set to demo-user)
3.  rio kubectl get -n (namespace) -o=json deploy/service

Results: demo-user is not set in the container / securityContext section

StrongMonkey:
Will be fixed in https://github.com/rancher/rio/pull/342.

PR #342, opened 2019-07-31
Fix user and group not passed in cli

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/163

Issue #351, opened 2019-08-05
Rio up bring zero services


StrongMonkey:
Version: v0.3.0
Steps to reproduce:
1. Create Riofile like this:
```
services:
  demo:
    image: nginx
    ports:
    - 80/http
    scale: 1-20
```
2. Run `rio up`
3. Service come up with zero scale.

harutyundermenjyan:
Also, the manual change of scaling with range e.g $NAMESPACE/$SERVICE=1-5 causes for downscaling to 0 scale.

StrongMonkey:
@harutyundermenjyan yes it will also be fixed in the pull request.

StrongMonkey:
Fixed in v0.3.1-rc1.

PR #352, opened 2019-08-05
Fix scale not set properly in CLI and Riofile

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/351

PR #353, opened 2019-08-05
Fix secretRef and configMapRef doesn't get pass through

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/354

PR #370, opened 2019-08-08
Remove "value" parameter from EnvVar when using config:// or secret:// as --env parameter

daniel-sullivan:
Removed "value" parameter from EnvVar passed onto K8S


StrongMonkey:
@daniel-sullivan Thanks for the contribution! LGTM

Issue #367, opened 2019-08-07
rio doesn't support cpu and memory based autoscalling


harutyundermenjyan:
Hello,
I think that cpu and memory based autoscaling still will be actual for a long time, so it will be very useful if they added in rio.
Or if it turns out difficult to implement, even let to work out the k8s hpa in some way for the service created by rio.

StrongMonkey:
Available to test in v0.4.0-rc1:
To create a service that doesn't control the scale of a deployment, set scale=-1:
`rio run --scale=-1 ${image}`

Find the deployment that it creates(should be in same name/namespace). You should be able to change the scale of the deployment and rio will not try to revert it back.

tfiduccia:
Version - 0.4.0-rc2 Verified implemented

PR #384, opened 2019-08-14
Allow other components to change scale if scale is -1

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/367

PR #442, opened 2019-09-16
Fix public domain ingress, delete unused module

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/444

PR #451, opened 2019-09-16
Update go mod tidy (and fix for Azure/go-autorest & azure-sdk-for-go)

daxmc99:
This is a temporary fix similar to that from https://github.com/coredns/coredns/pull/3258

> See https://github.com/coredns/coredns/pull/3209#issuecomment-525016357 for the previous issues, also see Azure/go-autorest#414 for some further background.
Basically go mod has some issue when multiple tags exist (as main module and submodule) within the same repo. That was fixed in go-autorest very recently, though azure-sdk-for-go has not been fully updated yet.
This fix is a temporarily one with replace fix. Once azure-sdk-for-go is updated then we could drop the replace fix

This fix can be dropped once our dependencies are updated to a more recent version of the `Azure/go-autorest`
Note: This only occurs with `go mod tidy` on go1.13


ibuildthecloud:
Hmm, still conflicting/faliing

PR #452, opened 2019-09-16
Ensure registry proxy is starting

cbron:
This fixes the _"no available registry endpoint: failed to do request: Head http://localhost:5442"_ error


StrongMonkey:
This change make sure that after we register controller, we are re-enqueue all objects to make sure it gets processed.

StrongMonkey:
https://github.com/rancher/rio/issues/453

cbron:
I'm surprised that `test_rio_app_endpoint` didn't catch this. Maybe just differing start times on local vs drone ?

StrongMonkey:
@cbron  Our test didn't test build functionality.

Issue #421, opened 2019-09-11
rio logs doesn't show timestamp


ibuildthecloud:

StrongMonkey:
The fix should be made in CLI to include timestamps in logs.

StrongMonkey:
Verified with v0.5.0-alpha.3. 
`rio -s logs -f -t rio-system/cert-manager` shows
```
cert-manager-75cccd99bd-g8r4s/cert-manager | 2019-09-24T18:59:04.885355661Z I0924 18:59:04.885164       1 controller.go:184] orders controller: syncing item 'rio-system/rio-wildcard-886902865'
cert-manager-75cccd99bd-g8r4s/cert-manager | 2019-09-24T18:59:04.887108767Z I0924 18:59:04.886948       1 controller.go:190] orders controller: Finished processing work item "rio-system/rio-wildcard-886902865"
```

PR #456, opened 2019-09-17
Timestamp in logs

cbron:
Issue https://github.com/rancher/rio/issues/421

First commit is just to do a clean `go mod vendor`.

Second commit adds the timestamp flag to the `rio logs` command.

```
$ rio logs --help
...
--timestamps, -t             Print the logs with timestamp
```

Examples: 

`rio logs --timestamps example`:
> example-8545778c56-6qfxd/linkerd-init  | 2019-09-17T15:21:56.1789494Z Chain OUTPUT (policy ACCEPT 0 packets, 0 bytes)

`rio logs example`:
> example-8545778c56-6qfxd/linkerd-init  | Chain OUTPUT (policy ACCEPT 0 packets, 0 bytes)

This can also lead to rather ugly double timestamps: 
> mine-8545778c56-6qfxd/linkerd-init  | 2019-09-17T15:21:56.1745533Z 2019/09/17 15:21:56 > iptables -t nat -vnL

### Testing

* `rio run --name example https://github.com/rancher/rio-demo`
* `rio logs example` will return normally
* `rio logs --timestamps example` will now show timestamp


cli/cmd/logs/logs.go
```diff
@@ -17,13 +17,14 @@ import (
...
-	N_Tail      int    `desc:"Number of recent lines of logs to print, -1 for all" default:"200"`
-	A_All       bool   `desc:"Include hidden or systems logs when logging"`
+	F_Follow            bool   `desc:"Follow log output"`
+	S_Since             string `desc:"Logs since a certain time, either duration (5s, 2m, 3h) or RFC3339"`
+	P_Previous          bool   `desc:"Print the logs for the previous instance of the container in a pod if it exists"`
+	C_Container         string `desc:"Print the logs of a specific container"`
+	R_Revision          string `desc:"Print the logs of a specific revision"`
+	N_Tail              int    `desc:"Number of recent lines of logs to print, -1 for all" default:"200"`
+	A_All               bool   `desc:"Include hidden or systems logs when logging"`
+	DT_DisableTimestamp bool   `desc:"Print the logs without timestamp"`
```
StrongMonkey: What does the flag looks like after the change? the way it is structured right now is : A_BC so that long flag is BC and short flag is A. But I don't know if short flag allows two letters. Worth to check.

cbron: ```   --disable-timestamp, --dt    Print the logs without timestamp```

Open to suggestions on better naming too

StrongMonkey: Ok. Looks good. Left to @ibuildthecloud for naming things. Personally I think it should be disabled by default?

cbron: I would agree, think it's a lot of data and rather ugly.

ibuildthecloud: I'd probably just follow what docker does 
```
-t, --timestamps     Show timestamps
```

ibuildthecloud: so off by default.

cbron:
Comments addressed. Still only look at most recent commit.

PR #473, opened 2019-09-20
WIP: Ginkgo poc

cbron:
https://github.com/rancher/rio/issues/486

[Ginkgo library](https://onsi.github.io/ginkgo/)

Run tests with: `ginkgo -p -v -tags validation ./tests_go/...`

`go test ./...` will not pickup our suite because of build tags

## Directory layout

* tests_go (will change this name later) 
  * testutil - handles all communal code, and setup/teardown
    * testutil.go
    * service.go - since go is typed we need to break up util code to be type specific
    * wait.go - handles the `waitFor x` stuff
  * validation - contains our suite, we can break it up into subfolders if necessary
    * suite_test.go - main file
    * run_test.go - test `rio run`
    * scale_test.go - test `rio scale` 
    * logs_test.go  - test `rio logs`


Dockerfile.dapper
```diff
@@ -11,6 +11,8 @@ RUN go get -d golang.org/x/lint/golint && \
     git -C /go/src/golang.org/x/lint/golint checkout -b current 06c8688daad7faa9da5a0c2f163a3d14aac986ca && \
     go install golang.org/x/lint/golint && \
     rm -rf /go/src /go/pkg
+RUN go get github.com/onsi/ginkgo/ginkgo && \
+    rm -rf /go/src /go/pkg
```
daxmc99: No need for a second RUN command, should be moved up

Issue #432, opened 2019-09-13
image: PATH should work


ibuildthecloud:
In a Riofile the following should work

```yaml
services:
  foo:
    image: ./blah
```

In this situation we should detect that ./blah is a path and then setup the `build` element appropriately.

daxmc99:
Should this work for remote repositories and local ones?

daxmc99:
This can be tested with https://github.com/daxmc99/rio-demo/tree/test2
Clone and run `rio up` in this directory

izaac:
@daxmc99 in `rio v0.6.0` I end up with the following after `rio up`:

```bash
➜  dax-rio-demo git:(test2) rio ps
NAME      IMAGE                                                                                         ENDPOINT   SCALE     APP       VERSION    WEIGHT    CREATED          DETAIL
nginx     default/nginx-nginx@sha256:56a6cd38367f6dac4cbf88e8fd60ee3be77fd34aa5e7b6a925e4c6cf9a38e6dc              0/1 0%    nginx     60fe91b3   100%      44 seconds ago   nginx: nginx(ErrImagePull/rpc error: code = Unknown desc = failed to pull and unpack image "localhost:5442/default/nginx-nginx@sha256:56a6cd38367f6dac4cbf88e8fd60ee3be77fd34aa5e7b6a925e4c6cf9a38e6dc": failed to resolve reference "localhost:5442/default/nginx-nginx@sha256:56a6cd38367f6dac4cbf88e8fd60ee3be77fd34aa5e7b6a925e4c6cf9a38e6dc": failed to do request: Head https://localhost:5442/v2/default/nginx-nginx/manifests/sha256:56a6cd38367f6dac4cbf88e8fd60ee3be77fd34aa5e7b6a925e4c6cf9a38e6dc: http: server gave HTTP response to HTTPS client)
```

This is a k3s v0.10.0 cluster hosted on Linode.
```bash 
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2-k3s.1", GitCommit:"b8b17ba55f20e590df507fce333dfee13ab438c6", GitTreeState:"clean", BuildDate:"2019-10-16T05:17Z", GoVersion:"go1.13.3", Compiler:"gc", Platform:"linux/amd64"}```

rancher-max:
Looks like this is still failing in v0.6.0-alpha.2.

Ran `rio up --name dax --branch test2 https://github.com/daxmc99/rio-demo`
[...]

PR #483, opened 2019-09-20
Add ability to provide imagePath in Riofile

daxmc99:
Addresses https://github.com/rancher/rio/issues/432

### DIFF
--- a/pkg/stack/stack.go
+++ b/pkg/stack/stack.go
@@ -2,8 +2,10 @@ package stack
 
 import (
 	"encoding/json"
+	"os"
 	"strings"
 
+	"github.com/pkg/errors"
 	riov1 "github.com/rancher/rio/pkg/apis/rio.cattle.io/v1"
 	v1 "github.com/rancher/rio/pkg/apis/rio.cattle.io/v1"
 	"github.com/rancher/rio/pkg/riofile"
@@ -94,6 +96,20 @@ func (s *Stack) GetImageBuilds() (map[string]riov1.ImageBuild, error) {
 				}
 
 				buildConfig[svc.Name] = *svc.Spec.Build
+			} else {
+				// check if the image points toward a local path
+				if strings.HasPrefix(svc.Spec.Image, "./") || strings.HasPrefix(svc.Spec.Image, "/") {
+					fileInfo, err := os.Stat(svc.Spec.Image)
+					if err != nil {
+						return nil, errors.Wrap(err, "error parsing image field")
+					}
+					if fileInfo.IsDir() {
+						buildConfig[svc.Name] = riov1.ImageBuild{BuildContext: svc.Spec.Image}
+						svc.Spec.Image = ""
+					} else {
+						return nil, errors.Wrap(err, "image field is not a directory")
+					}
+				}
 			}
 		}
 	}


pkg/stack/stack.go
```diff
@@ -94,6 +97,21 @@ func (s *Stack) GetImageBuilds() (map[string]riov1.ImageBuild, error) {
 				}
 
 				buildConfig[svc.Name] = *svc.Spec.Build
+			} else {
+				// use docker reference parser to determine if its a valid image reference
+				_, err = reference.ParseAnyReference(svc.Spec.Image)
```
StrongMonkey: Any reason we need parse the image? Can we assume that if it begins with `./` then use that as the local reference path?

daxmc99: I think path such as `/Users/repo/project/` could also be used? Not guaranteed to be a relative path. But the docker image parsing might be overkill beyond `os.Stat`

cbron: Looks like this also accepts a commit sha, but that would fail out below. Guessing the benefit in using it is that an invalid image string will fail before the `os.stat` call, even if the file exists.

pkg/stack/stack.go
```diff
@@ -94,6 +97,21 @@ func (s *Stack) GetImageBuilds() (map[string]riov1.ImageBuild, error) {
...
+				if err != nil { //fallback and see if its directory
+					fileInfo, err := os.Stat(svc.Spec.Image)
+					if err != nil {
+						return nil, errors.Wrap(err, "error parsing image field")
+					}
+					if fileInfo.IsDir() {
+						buildConfig[svc.Name] = riov1.ImageBuild{BuildContext: svc.Spec.Image}
+						svc.Spec.Image = ""
+					} else {
+						return nil, errors.Wrap(err, "image field is not a directory")
```
cbron: Is there a reason it has to be a dir, can't I just have 5 named dockerfiles in this dir ? `Dockerfile.agent`, `Dockerfile.web`, `Dockerfile.db`  ...

pkg/stack/stack.go
daxmc99: @cbron if its an image SHA than we would essentially do nothing here, which the the correct and current behavior (you are technically referencing an image, not something to build)

pkg/stack/stack.go
daxmc99: This sounds like a design question and my answer is leaning towards no. 
1. If you give us a directory where we should have had an `image` then
2. We will do some _minor_ inferencing and try to build the Dockerfile we find in there. 
3. We are mimicking the behavior of Dockerbuild for a simple flow https://docs.docker.com/engine/reference/commandline/build/#build-with-path
- More advanced flows should be defining multiple images within the Riofile construct

cbron: Yea Riofile should work for now or they can just make dirs. If someone requests we can make a future enhancement.

StrongMonkey: Dockerfile name can be passed in build args though. We have `Dockerfile` field to specify the name of Dockerfile. It defaults to `Dockerfile`.

pkg/stack/stack.go
```diff
@@ -94,6 +96,20 @@ func (s *Stack) GetImageBuilds() (map[string]riov1.ImageBuild, error) {
 				}
 
 				buildConfig[svc.Name] = *svc.Spec.Build
+			} else {
+				// check if the image points toward a local path
+				if strings.HasPrefix(svc.Spec.Image, "./") || strings.HasPrefix(svc.Spec.Image, "/") {
```
daxmc99: @StrongMonkey only checking prefixs now

Issue #481, opened 2019-09-20
Publicdomain failed to request letsencrypt certificates if pointing to router


StrongMonkey:
Version: v0.5.0-alpha.2

Steps to reproduce:

Register foo.bar domain to router
`Rio domain register foo.bar router`

Access https://foo.bar
Letsencrypts certs fail to provision.

cbron:
Testing instructions: 

1. create a router to point to a service
2. create a public domain, change the dns record cname point to cluster domain in rio. (xxx.on-rio.io showing up in rio info)
3. rio domain register $domain $name_of_router
4. access that domain, it should be trusted as we requested a certs for that public domain.

StrongMonkey:
Available to test in v0.5.0-alpha.3.

StrongMonkey:
@cbron I use the follow instruction to test:
1. Create a GKE cluster and install rio. GKE will expose a loadbalancer that has a public IP on it.
2. Create a service service-foo and router router-foo following the docs.
3. Create two custom domains(I did this in DO space) to map CNAME record to rio cluster domain.
4. Use `rio domain register` to register domain to service-foo and router-foo.
5. Make sure you can access your services and routers by that public domain, and also certs are provisioned for those domains.

cbron:
Working correctly in alpha3

```
$ rio run --name demo1 https://github.com/rancher/rio-demo
$ rio ps

NAME            CREATED          ENDPOINT                                 REVISIONS   SCALE     WEIGHT    DETAIL
default/demo1   3 minutes ago   https://demo1-default.zsak9z.on-rio.io   v0          1         100%

$ rio route add x.default to default/demo1
$ rio route

NAME          URL                                    OPTS      ACTION    TARGET
default/x     https://x-default.zsak9z.on-rio.io               to        demo1,port=80

$ rio domain register caleb.do.rancher.space default/x
$ rio domain register caleb2.do.rancher.space default/x
$ rio domain

NAME                      DOMAIN                   TARGET
default/caleb-do--7bff2   caleb.do.rancher.space   default/x
default/caleb2-do-2b467   caleb2.do.rancher.space   default/x

$ curl https://caleb.do.rancher.space
Hello World

$ openssl s_client -connect caleb.do.rancher.space:443

Certificate chain
 0 s:/CN=*.zsak9z.on-rio.io
   i:/C=US/O=Let's Encrypt/CN=Let's Encrypt Authority X3
....
```

PR #484, opened 2019-09-20
Fix publicdomain failing to create certificates

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/481

modules/gateway/controllers/publicdomain/populate/destinationrule.go
```diff
@@ -17,7 +17,7 @@ func DestionationRule(pd *adminv1.PublicDomain, systemNamespace string, os *obje
 			Host: fmt.Sprintf("%s.%s.svc.cluster.local", name, systemNamespace),
```
daxmc99: Typo
```suggestion
	func DestinationRule(pd *adminv1.PublicDomain, systemNamespace string, os *objectset.ObjectSet)
```

modules/gateway/controllers/publicdomain/populate/destinationrule.go
```diff
@@ -17,7 +17,7 @@ func DestionationRule(pd *adminv1.PublicDomain, systemNamespace string, os *obje
...
 			Subsets: []v1alpha3.Subset{
 				{
-					Name: "latest",
```
daxmc99: Does this name change do anything?

modules/gateway/controllers/service/populate/virtualservice.go
```diff
@@ -110,7 +110,7 @@ func newRoute(systemNamespace, externalGW string, published bool, portBinding v1
 			ns = svc.Namespace
 		}
 		host := dests[0].Host
-		if dests[0].Subset != "" {
+		if dests[0].Subset != "" && dests[0].Subset != "cm-acme" {
```
cbron: Should probably be a constant if we are using in multiple places

modules/gateway/controllers/publicdomain/populate/destinationrule.go
StrongMonkey: It doesn't, but would like a special name we can tell so that we don't program in the following place 
```
if dests[0].Subset != "" && dests[0].Subset != "cm-acme" {
```

modules/gateway/controllers/service/populate/virtualservice.go
StrongMonkey: :+1:

Issue #485, opened 2019-09-21
Promoting a new revision the parent service URL still redirects to the old revision


izaac:
Steps:

- `rio run --scale 2 -p 80 -n appone  sangeetha/testupgrade:v2`
   - Got `appone:v0` created and the parent ENDPOINT for the service.
- `rio stage --image sangeetha/testupgrade:v3 default/appone`
  - Got a new revision for appone with its own endpoint and WEIGHT =0 
  - Parent ENDPOINT still same, `curl`ing the parent endpoint still gets `appone:v0`
- `rio promote default/appone:<new revision>`
  - the new revision starts rolling out, increasing its weight
- New appone revision gets 100% WEIGHT, but parent ENDPOINT still reaching revision `v0` `testupgrade:v2` contents..

GKE Cluster v1.14.6-gke.1 
Default install `linkerd`

StrongMonkey:
Available to test in v0.5.0-alpha.3.

izaac:
This is working now in `v0.5.0-alpha.3`. 

- Using `istio` mesh mode
- Using `linkerd` mesh mode

PR #488, opened 2019-09-21
Fix traffic split issue

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/485

modules/gateway/controllers/routeset/populate/virtualservice.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/rancher/mapper/slice"
+
```
daxmc99: nit: formatting

modules/gateway/controllers/routeset/populate/virtualservice.go
```diff
@@ -66,16 +68,14 @@ func virtualServiceFromRoutesets(systemNamespace string, clusterDomain *projectv
 				port = int(*dest.Port)
 			}
 
-			if i == len(routeSpec.To)-1 && constants.ServiceMeshMode == constants.ServiceMeshModeLinkerd {
-				host := dest.Service
-				if dest.Revision != "" {
-					host = host + "-" + dest.Revision
+			if constants.ServiceMeshMode == constants.ServiceMeshModeLinkerd {
+				httpRoute.Headers.Request.Set["l5d-dst-override"] = fmt.Sprintf("%s-%v.%s.svc.cluster.local:%v", routeSet.Name, index, routeSet.Namespace, port)
```
daxmc99: I feel like we need to mention something about https://github.com/linkerd/linkerd2-proxy/pull/173
Or document what `l5d-dst-override` does. I'm even fine with a link the linkerd2 PR. I feel like this is hard to understand without some more context with a comment

StrongMonkey:
This pr is to fix two things:
1. TrafficSplit on routeset should work. This is controlled by overriding header in virtualservice with 
routeName-${index}.namespace.svc.cluster.local. (Index is the number of route in router),  for each index a trafficSplit is created. 
2. TrafficSplit on app(aggregation of services) is done by overriding header in virtual service with ${app_name}.namespace.svc.cluster.local. That's why we need aggregated  flag to control whether we write version or no.

modules/gateway/controllers/service/populate/virtualservice.go
```diff
@@ -264,7 +264,7 @@ func isProtocolSupported(protocol v1.Protocol) bool {
 }
 
 func VirtualServiceFromSpec(aggregated bool, systemNamespace string, name, namespace string, clusterDomain *projectv1.ClusterDomain, service *v1.Service, dests ...Dest) *v1alpha3.VirtualService {
```
daxmc99: Could we add a comment to define what `aggregated` does?

daxmc99: My understanding is that aggregated determines if its the route to a revision versus the "app" itself.

modules/gateway/controllers/service/populate/virtualservice.go
```diff
@@ -100,7 +100,7 @@ func httpRoutes(systemNamespace string, service *v1.Service, dests []Dest) ([]v1
 
 	return result, external
 }
-func newRoute(systemNamespace, externalGW string, published bool, portBinding v1.ContainerPort, dests []Dest, appendHTTPS bool, autoscale bool, svc *v1.Service) (string, v1alpha3.HTTPRoute) {
+func newRoute(aggregated bool, systemNamespace, externalGW string, published bool, portBinding v1.ContainerPort, dests []Dest, appendHTTPS bool, autoscale bool, svc *v1.Service) (string, v1alpha3.HTTPRoute) {
```
daxmc99: Maybe the explanation would be better here?

StrongMonkey:
@daxmc99 Please review, comments has been addressed.

Issue #482, opened 2019-09-20
hostport mode install is stuck using v0.5.0-alpha.2


izaac:
GKE cluster. v1.14.6-gke.1 Provisioned with Rancher.

`rio install --mode hostport`

Installation gets stuck for a long time:

```
Deploying Rio control plane....
- Waiting for all the system components to be up. Not ready: [gateway]
```

Logs collected:

`rio systemlogs`
`rio-controller | E0920 21:51:31.959229       1 controller.go:117] error syncing 'rio-system/cluster-domain': handler rdns: failed to create domain, requeuing`

`kubectl get pods --all-namespaces`

```
rio-system      registry-proxy-kjfm5                             0/1     ContainerCreating   0          13m
rio-system      registry-proxy-lvb89                             0/1     ContainerCreating   0          13m
rio-system      registry-proxy-wvf79                             0/1     ContainerCreating   0          13m
```

StrongMonkey:
Available to test in v0.5.0-alpha.3

izaac:
I did validations with `v0.5.0-alpha.3` in RKE and GKE clusters. using istio and linkerd mesh modes.

Looking good, it didn't took too long to setup rio this time using `hostport`

PR #489, opened 2019-09-22
Fix hostport install stucking at gateway

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/482

Issue #486, opened 2019-09-21
Convert integration tests to Go


cbron:
To keep everything in one language, we are going to migrate our Python testing stack to go. 

We are also going to move to a BDD style of testing to better integrate with requirements and make things more obvious to wider community.

PR #493, opened 2019-09-23
WIP: Spec poc

cbron:
https://github.com/rancher/rio/issues/486

[Spec library](https://github.com/sclevine/spec)

This has the same tests as the ginkgo POC. Primary difference is that it uses spec framework instead of ginkgo, and also uses testify instead of gomega. It also doesn't use build flags but rather command-line flags, which afaik is not possible with ginkgo because lack of access to `testing.M`. 

Run tests with: `go test -v ./tests_go/... -validation-tests`

The standard `go test ./...` will not pickup our suite because CLI flag. 

## Directory layout

(Same as ginkgo pr) 

* tests_go (will change this name later) 
  * testutil - handles all communal code, and setup/teardown
    * testutil.go
    * service.go - since go is typed we need to break up util code to be type specific
    * wait.go - handles the `waitFor x` stuff
  * validation - contains our suite, we can break it up into subfolders if necessary
    * suite_test.go - main file
    * run_test.go - test `rio run`
    * scale_test.go - test `rio scale` 
    * logs_test.go  - test `rio logs`


cbron:
Review requested for spec vs gingko, overall approach, and layout mainly. Not so much code specifics, that can be cleaned up later.

StrongMonkey:
Wow only 2k lines of changes :+1:

cbron:
Yea this was rush job since it's just POC. Whichever one we go with these PRs will get thrown away and we'll just copy over code to a legitimate/clean PR.

Issue #479, opened 2019-09-20
Autoscaling is not setting the scale up in service with v0.5.0-alpha.2


izaac:
Create a cluster, I used a GKE cluster created with Rancher.

create a service:

`rio run -p 80 -n version3 sangeetha/testupgrade:v3`

hit the http://<service>:9080 plain http service port with `hey`

I used the following parameters:

`hey -z 3m -c 100 http://<service>:9080`

`rio ps` is always showing SCALE = 1.

Collected logs in istio mode:

`autoscaler-5dfb757594-7s7t4/autoscaler | {"level":"info","ts":"2019-09-20T20:51:14.122Z","logger":"autoscaler.kpa-class-podautoscaler-controller.event-broadcaster","caller":"record/event.go:221","msg":"Event(v1.ObjectReference{Kind:\"PodAutoscaler\", Namespace:\"default\", Name:\"version1\", UID:\"f3307250-dbe1-11e9-9ddd-42010a8000b6\", APIVersion:\"autoscaling.internal.knative.dev/v1alpha1\", ResourceVersion:\"37669\", FieldPath:\"\"}): type: 'Warning' reason: 'InternalError' error reconciling SKS: error updating SKS version1: Operation cannot be fulfilled on serverlessservices.networking.internal.knative.dev \"version1\": the object has been modified; please apply your changes to the latest version and try again","knative.dev/controller":"kpa-class-podautoscaler-controller"}`

izaac:
same behavior with `linkerd` mode

StrongMonkey:
Available to test in v0.5.0-alpha.3. Also @izaac can you test with my image `strongmonkey1992/autoscale:v0`

izaac:
@StrongMonkey using `v0.5.0-alpha.3` Autoscaling seems to work with your image.


```
NAME                 CREATED          ENDPOINT                                           REVISIONS   SCALE     WEIGHT    DETAIL
default/autoscale1   3 minutes ago    https://autoscale1-default.738yh7.on-rio.io:9443   v0          7         100%
```

<img width="1387" alt="Screen Shot 2019-09-24 at 1 48 40 PM" src="https://user-images.githubusercontent.com/27659/65549329-29f60c00-ded2-11e9-803b-757623be2a2d.png">

izaac:
Both `istio` and `linkerd`.

PR #501, opened 2019-09-24
Update with linkerd metric fix, fix scale to zero bug

StrongMonkey:
Fix linkerd header to include activator as header override so that
when service is scaled to zero it sends traffic to activator service


StrongMonkey:
https://github.com/rancher/rio/issues/479

modules/gateway/controllers/service/populate/virtualservice.go
```diff
@@ -113,15 +113,22 @@ func newRoute(systemNamespace, externalGW string, published bool, portBinding v1
...
+		if autoscale && deployIsZero(svc) {
+			host = "activator"
+			ns = systemNamespace
+			port = 8012
+		}
 		route.Headers = &v1alpha3.Headers{
 			Request: &v1alpha3.HeaderOperations{
 				Set: map[string]string{
-					"l5d-dst-override": fmt.Sprintf("%v.%s.svc.cluster.local:%v", host, ns, portBinding.Port),
+					"l5d-dst-override": fmt.Sprintf("%v.%s.svc.cluster.local:%v", host, ns, port),
```
cbron: Should we turn `l5d-dst-override` into a const too ? That also brings up "l5d-remote-ip" and "l5d-server-id". I think it would be better to at least do the override, a const can avoid typos, also easier to search or change all uses.

Issue #491, opened 2019-09-23
rio SUBCOMMAND --help should work when there is no server available


ibuildthecloud:
If I type `rio promote --help` or any subcommand that there is no valiid kubeconfig we show the global help and not the help for this subcommand.

StrongMonkey:
Tested with master build, works as expected.

PR #502, opened 2019-09-24
Print subcommand help 

daxmc99:
This commit add support for printing out the help for a specific
subcommand when a commands matching the form rio SUBCOMMAND --help
are entered.

Addresses https://github.com/rancher/rio/issues/491

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -230,7 +230,15 @@ func main() {
 	}
 	app.ExitErrHandler = func(context *cli.Context, err error) {
 		if err != nil && strings.Contains(err.Error(), clicontext.ErrNoConfig.Error()) {
-			printConfigUsage()
+			if context.Args().Get(context.NArg()-1) == "--help" {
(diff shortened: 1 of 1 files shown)


cli/main.go
```diff
@@ -230,7 +230,15 @@ func main() {
 	}
 	app.ExitErrHandler = func(context *cli.Context, err error) {
 		if err != nil && strings.Contains(err.Error(), clicontext.ErrNoConfig.Error()) {
-			printConfigUsage()
+			if context.Args().Get(context.NArg()-1) == "--help" {
+				err = cli.ShowCommandHelp(context, context.Args().First())
+				if err != nil {
+					cli.HandleExitCoder(err)
+				}
```
cbron: Minor, but I feel like in this case (calling `--help` and it works) the exit code should be 0 since nothing failed.

daxmc99: Fixed!

Issue #263, opened 2019-05-23
"v0 NotReady; v0 waiting on build" when following Quick Start docs


maoo:
Environment info:
```
~ » rio info
Rio Version: v0.1.1-rc1 (cdb75cf1)
Rio CLI Version: v0.1.1-rc1 (cdb75cf1)
Cluster Domain: dwxqoo.on-rio.io
Cluster Domain IPs: 127.0.0.1
System Namespace: rio-system

~ » kubectl version
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.11", GitCommit:"637c7e288581ee40ab4ca210618a89a555b6e7e9", GitTreeState:"clean", BuildDate:"2018-11-26T14:38:32Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.11", GitCommit:"637c7e288581ee40ab4ca210618a89a555b6e7e9", GitTreeState:"clean", BuildDate:"2018-11-26T14:25:46Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```

I've followed [Quick Start instructions](https://github.com/rancher/rio#quick-start), but when running `rio ps` I get the following:

```
NAME                      CREATED         ENDPOINT                                                REVISIONS   SCALE     WEIGHT    DETAIL
default/pensive-mclean2   8 minutes ago   https://pensive-mclean2-default.dwxqoo.on-rio.io:9443   v0          0/1       100%      v0 NotReady; v0 waiting on build
```

When I try to access the URL, I get `no healthy upstream`:
```
~ » curl -k https://pensive-mclean2-v0-default.dwxqoo.on-rio.io:9443/
no healthy upstream%
```

plunkettscott:
I was also getting this, however, I came back about 40 hours later and now all I see is the following:

NAME                         CREATED        ENDPOINT                                                   REVISIONS   SCALE     WEIGHT    DETAIL
default/infallible-clarke2   47 hours ago   https://infallible-clarke2-default.hdyhdj.on-rio.io:9443   v0          0/1       100%      v0 NotReady

StrongMonkey:
can you post systemlogs here by running the command `rio systemlogs`? Also can you make sure all the system pods are up and running? We fixed an issue related to registry proxy not starting in rc2 and it might be related to your issue.

maoo:
[...]

PR #503, opened 2019-09-24
Only deploy socat with hostNetwork to proxy registry

StrongMonkey:
For local registry we previously use hostport to expose registry service
to localhost:5442. Since hostports are wonky and won't work in every k8s
distro, switching to socat container with hostNetwork: true would be the
best solution


StrongMonkey:
Hopefully this pr should address the issue https://github.com/rancher/rio/issues/263 since we are defaulting to use hostNetwork instead of hostport and it is more likely to work.

PR #506, opened 2019-09-24
Fix up ImageBuild

ibuildthecloud:


pkg/apis/rio.cattle.io/v1/image.go
```diff
@@ -45,23 +49,21 @@ type ImageBuildSpec struct {
...
-	BuildTimeout *metav1.Duration
-
-	// ServiceName The service to update when the build succeeds or fails.  If blank do not update any service.
-	ServiceName string `json:"serviceName,omitempty"`
+	// Timeout describes how long the build can run
+	Timeout *metav1.Duration
 }
 
 type ImageBuildStatus struct {
+	ImageName string
```
daxmc99: Should this contain a field that determines if the image is done building?

StrongMonkey: @daxmc99 I don't think darren care too much about fields in status. So add those fields if you need. Those are mostly internal types that are consumed by controllers.

Issue #507, opened 2019-09-24
Grafana installation is stuck in lite install with mesh mode istio


izaac:
**Describe the bug**
Install with `rio install --lite --mesh-mode istio`

I've waited for the installation to finish for 20 minutes.

`/ Waiting for all the system components to be up. Not ready: [grafana]`

**To Reproduce**
1. Setup a GKE cluster I used v1.14.6-gke.1 as k8s version.
2. Install rio using `rio install --lite --mesh-mode istio`

**Expected behavior**

Finish the install of all rio lite mode features using `--lite` and `--mesh-mode istio`

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
v1.14.6-gke.1
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.5.0-alpha.3 (cdb631d1)
Rio CLI Version: v0.5.0-alpha.3 (cdb631d1)
Cluster Domain: <domain>.on-rio.io
Cluster Domain IPs: X.X.X.X
System Namespace: rio-system

System Components:
activator status: running
autoscaler status: running
build-controller status: running
buildkitd-dev status: running
cert-manager status: running
controller status: running
gateway status: running
istio-citadel status: running
istio-pilot status: running
istio-sidecar-injector status: running
istio-telemetry status: running
kiali status: running
prometheus status: running
registry status: running
webhook status: running
```
**Additional context**
`rio systemlog`
output:
```
No output
```

StrongMonkey:
@izaac available to test in v0.5.0-alpha.4

izaac:
@StrongMonkey 

Now after this change using v0.5.0-alpha.4:
`rio install --mesh-mode istio`

```
Deploying Rio control plane....
\ Waiting for all the system components to be up. Not ready: [mixer]
```

The installation just sits there. I gave up after 12 minutes.

These other installation combinations worked: 

`rio install --lite --mesh-mode istio` ✅ 
`rio install --lite --mesh-mode linkerd` ✅ 
`rio install ` ✅

izaac:
Above installation combinations are now working as expected using v0.5.0-alpha.4 🥇 

Closing.

PR #508, opened 2019-09-24
Fix lite install being broken in istio mesh mode

StrongMonkey:
Fix lite install paramaters being broken due to the wrong parameters and
template patterns.


StrongMonkey:
https://github.com/rancher/rio/issues/507

cli/cmd/install/install.go
```diff
@@ -89,7 +89,7 @@ var (
 	IstioGrafana           = "grafana"
 	IstioCitadel           = "istio-citadel"
 	IstioPilot             = "istio-pilot"
-	IstioTelemetry         = "istio-telemetry"
+	IstioTelemetry         = "mixer"
```
daxmc99: Same as above, does it make sense to use constants here?

cli/cmd/install/install.go
```diff
@@ -114,7 +85,7 @@ type Install struct {
 	HTTPProxy       string   `desc:"Set HTTP_PROXY environment variable for control plane"`
 	Yaml            bool     `desc:"Only print out k8s yaml manifest"`
 	Check           bool     `desc:"Only check status, don't deploy controller"`
-	Lite            bool     `desc:"Only install lite version of Rio(monitoring will be disabled, will be ignored if --disable-features is set)"`
+	Lite            bool     `desc:"Only install lite version of Rio istio install(only works if mesh-mode is istio, monitoring will be disabled, will be ignored if --disable-features is set)"`
```
izaac: do we need to prevent the users from `rio install --lite --mesh-mode linkerd` and `rio install --lite` or will this silently install "full" when mesh mode is not `istio` ? @StrongMonkey

StrongMonkey:
@daxmc99 now with more constants, also remove used constants.

cli/cmd/install/install.go
StrongMonkey: @izaac it will ignore lite if using linkerd and install full linkerd. Let’s keep it right now as cli may get changed after cli review.

Issue #525, opened 2019-09-27
Rio doesn't work in eks 1.13


StrongMonkey:
Install Rio v0.5.0-rc1 in EKS

`rio systemlogs` will show the following error

```
E0926 17:10:04.138471    4161 reflector.go:126] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Failed to list *v1beta1.Ingress: the server could not find the requested resource
E0926 17:10:04.384964    4161 controller.go:117] error syncing 'rio-system/registry': handler ingress-app-object-controller: the server could not find the requested resource, handler routing-serviceset-object-controller: the server could not find the requested resource, requeuing
E0926 17:10:05.289905    4161 reflector.go:126] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Failed to list *v1beta1.Ingress: the server could not find the requested resource
E0926 17:10:06.446585    4161 reflector.go:126] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Failed to list *v1beta1.Ingress: the server could not find the requested resource
E0926 17:10:07.615544    4161 reflector.go:126] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Failed to list *v1beta1.Ingress: the server could not find the requested resource
E0926 17:10:08.782817    4161 reflector.go:126] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Failed to list *v1beta1.Ingress: the server could not find the requested resource
E0926 17:10:09.971205    4161 reflector.go:126] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Failed to list *v1beta1.Ingress: the server could not find the requested resource
E0926 17:10:11.129773    4161 reflector.go:126] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Failed to list *v1beta1.Ingress: the server could not find the requested resource
[...]

PR #526, opened 2019-09-27
Switch ingress to extension api

StrongMonkey:


StrongMonkey:
Need https://github.com/rancher/wrangler-api/pull/7

StrongMonkey:
This is to fix https://github.com/rancher/rio/issues/525.

ibuildthecloud:
hellz no?

ibuildthecloud:
okay.... I'll accept it.

StrongMonkey:
@daxmc99 @cbron Can you review this pr? I just found out that ingress for networking api won't work well with kubernetes 1.13 or below.

From 1.14 changelogs https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md
```
Ingress resources are now available via networking.k8s.io/v1beta1. Ingress resources in extensions/v1beta1 are deprecated and will no longer be served in v1.18. Existing persisted data is available via the new API group/version (#74057, @liggitt)
```

PR #533, opened 2019-09-27
Move to go testing

cbron:
Issue: https://github.com/rancher/rio/issues/486

In order to onboard devs and the community faster we are unifying
the codebase into go. We are also moving towards BDD style specs so
that the tests are more human readable, and our current suite (and its
gaps) are more immediately obvious.

See readme in changes for more info.


go.sum
```diff
@@ -381,6 +381,10 @@ github.com/rogpeppe/fastuuid v0.0.0-20150106093220-6724a57986af/go.mod h1:XWv6So
 github.com/rogpeppe/go-charset v0.0.0-20180617210344-2471d30d28b4/go.mod h1:qgYeAmZ5ZIpBWTGllZSQnw97Dj+woV0toclVaRGI8pc=
 github.com/russross/blackfriday v1.5.2 h1:HyvC0ARfnZBqnXwABFeSZHpKvJHJJfPz81GNueLj0oo=
 github.com/russross/blackfriday v1.5.2/go.mod h1:JO/DiYxRf+HjHt06OyowR9PTA263kcR/rfWxYHBV53g=
+github.com/sclevine/spec v1.2.0 h1:1Jwdf9jSfDl9NVmt8ndHqbTZ7XCCPbh1jI3hkDBHVYA=
+github.com/sclevine/spec v1.2.0/go.mod h1:W4J29eT/Kzv7/b9IWLB055Z+qvVC9vt0Arko24q7p+U=
```
daxmc99: did you run `go mod tidy`?

scripts/test
```diff
@@ -18,6 +18,8 @@ if [[ ${ARCH} == amd64 ]]; then
     rio-controller >/dev/null 2>&1 &
     rio install --check >/dev/null 2>&1
 
+    go test -v ./tests/integration/... --integration-tests
```
daxmc99: Does with work with `-race`

tests/integration/externalservice_test.go
```diff
@@ -0,0 +1,33 @@
+package integration
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/rancher/rio/tests/testutil"
+
+	"github.com/sclevine/spec"
```
daxmc99: nit: formatting

tests/integration/weight_test.go
```diff
@@ -0,0 +1,37 @@
+package integration
+
+import (
+	"testing"
+
+	"github.com/rancher/rio/tests/testutil"
```
daxmc99: Hmm, so should we do imports similar to rancher? 
std lib imports are grouped and all others are grouped in another block? 
I think we should

scripts/test
cbron: added

tests/integration/weight_test.go
cbron: Switched to the 3 groups: stdlib, external, and internal

PR #537, opened 2019-10-01
README update: route insert is a flag on add, not standalone

cbron:
Issue: https://github.com/rancher/rio/issues/512


cbron:
@vincent99 issue created for that convo, we can revisit at a later date.

Issue #545, opened 2019-10-02
Scale is not going above 10 when a range is given


rancher-max:
**Describe the bug**
Service is not scaling above 10 pods when a range above 10 is given. Upon inspection, the spec.maxScale and spec.minScale values are properly set in the CRD, but the actual scale and pod count only goes to 10.

**To Reproduce**
1. `rio run --scale=1-20 -n autoscale strongmonkey1992/autoscale:v0`
    - can set the scale to as low as you want and anything above 10. (validated with `scale=9-11`)
2. `rio inspect autoscale`
    - shows `maxScale` and `minScale` values in the resulting yaml as what they were set to in the run command.
3. `hey -z 3m -c 200 http://autoscale-default.<cluster domain>:9080`
    - while this is running, `watch rio ps` and should see the scale on the default/autoscale service increasing. With the amount of Requests per Second, I would expect this value to go above 10, but it does not.

**Expected behavior**
The SCALE should go above 10 at some point when doing autoscaling with a given range.

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-19T13:57:45Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.6-gke.2", GitCommit:"c9de33b5439df6e206d7ba646787c6ace92d737b", GitTreeState:"clean", BuildDate:"2019-09-06T18:30:33Z", GoVersion:"go1.12.9b4", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.5.0 (303f3652)
Rio CLI Version: v0.5.0 (303f3652)
Cluster Domain: x.on-rio.io
Cluster Domain IPs: x.x.x.x
System Namespace: rio-system

System Components:
activator status: running
autoscaler status: running
build-controller status: running
buildkitd-dev status: running
cert-manager status: running
controller status: running
gateway status: running
istio-pilot status: running
registry status: running
webhook status: running
```
**Additional context**
`rio systemlogs`
output:
There are a lot of similar attempts here, but here's the results after running the above steps:
```
[...]

PR #546, opened 2019-10-03
Bump serving max-scale-up-rate to allow scaling

cbron:
Issue: https://github.com/rancher/rio/issues/545

Currently this setting is acting like a max rather than max-rate, services never scale past this number. See issue for more info.

This will bring us inline with the upstream setting: https://github.com/knative/serving/pull/4527/commits/4b0c4ffd0ff0bf6b2a5912960ca4c8c0ed2902ec

It will allow the autoscaler to scale as necessary, but there won't be a "governor" to how fast it can scale. The standard service maxScale will override this, that defaults to 10 still.


stacks/rio-autoscaler-stack.yaml
```diff
@@ -1,7 +1,7 @@
 configs:
   config-autoscaler:
     enable-scale-to-zero: "true"
-    max-scale-up-rate: "10"
+    max-scale-up-rate: "1000"
```
daxmc99: The knative fix looked like a float, any reason we aren't using a float here?

cbron: Nope, we can. They specifically do float to show that it's a rate, so you can do like `1.1` to barely scale, but they've had it both ways in past.

Issue #378, opened 2019-08-08
[EPIC] - RBAC/Security


davidnuzik:
This issue is for tracking efforts around RBAC and security with relation to permissions model, etc.

https://github.com/rancher/rio/issues/298

davidnuzik:
Sync with Daishan for help validating this.

davidnuzik:
@davidnuzik has documentation requirements; needs docs issue.

rancher-max:
Issue 618 mentioned above is related to Rancher, so this is working as expected for Rio in v0.6.0-alpha.2 as 630 has been resolved.

PR #548, opened 2019-10-03
Auth Validating Webhook

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/378

pkg/server/startup.go
```diff
@@ -58,12 +60,36 @@ func Startup(ctx context.Context, systemNamespace, kubeConfig string) error {
...
+		constants.DevMode = "true"
+		controllerStack := stack.NewSystemStack(rioContext.Apply, systemNamespace, "rio-controller")
+		answer := map[string]string{
+			"NAMESPACE": systemNamespace,
+		}
+		if err := controllerStack.Deploy(answer); err != nil {
+			return err
+		}
+		webhookPort = ":7443"
+	}
```
cbron: So this block checks to see if rio-controller exists, and if it doesn't we know it's dev mode, but then if thats true it deploys rio-controller anyway ? Doesn't that conflict with the local dev controller ?

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,372 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/authentication/user"
+	request2 "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/client-go/tools/record"
+	rbacregistryvalidation "k8s.io/kubernetes/pkg/registry/rbac/validation"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	authWebhook = "auth-webhook"
```
cbron: Is this used ? If so I feel like we should use other constants instead.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,372 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+
+	if constants.DevMode != "" {
+		validatingWebhook.Webhooks[0].ClientConfig = v1beta1.WebhookClientConfig{
+			URL:      &[]string{fmt.Sprintf("https://%s:7443", devHost)}[0],
```
cbron: Perhaps make 7443 a constant, since we set it in multiple packages as string.

pkg/webhook/webhook.go
```diff
@@ -0,0 +1,372 @@
...
+							APIVersions: []string{"v1"},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	if constants.DevMode != "" {
+		validatingWebhook.Webhooks[0].ClientConfig = v1beta1.WebhookClientConfig{
```
cbron: If we are in dev mode, can we just empty the `rule.resources`, making everything open ?

cbron: Follow up question, is this going to be a feature flag ?

go.mod
```diff
@@ -9,26 +9,26 @@ replace (
 	github.com/jaguilar/vt100 => github.com/tonistiigi/vt100 v0.0.0-20190402012908-ad4c4a574305
 	github.com/jetstack/cert-manager => github.com/rancher/cert-manager v0.7.0-rio.1
 	github.com/knative/pkg => github.com/rancher/pkg v0.0.0-20190514055449-b30ab9de040e
+	github.com/linkerd/linkerd2 => github.com/StrongMonkey/linkerd2 v0.0.0-20191003215613-774744e91443
```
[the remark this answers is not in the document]
StrongMonkey: Actually....this includes the fix I need for linkerd2 to make it compile..

pkg/server/startup.go
StrongMonkey: In this rio-controller only if you send `RUN_CONTROLLER=true` as an answer so that it can deploy rio-controller. https://github.com/rancher/rio/pull/548/files/25d50f0ca2e14a5c63907169c206d0c469922698#diff-fc011507dab2bf2f5d8b5041a7a4d91cR278

pkg/webhook/webhook.go
StrongMonkey: I think we should always enable in dev mode in case we need to test this function, but webhook actually will allow everything if you have cluster-admin for privilege. Open to discussion for feature flag...

go.mod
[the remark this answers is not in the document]
cbron: Ahh, so it's not for local dev. Should we move too rancher org ?

pkg/server/startup.go
cbron: Ok, I think this will be solved by our feature flag convo.

go.mod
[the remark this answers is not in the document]
StrongMonkey: Yeah we should move this to rancher org eventually

tests/integration/rbac_test.go
```diff
@@ -0,0 +1,122 @@
...
+			testService.Kubeconfig = standardUser.GetKubeconfig()
+			testService.Create(t, "--permission", "list rio.cattle.io/services", "--disable-service-mesh", "nginx")
+		})
+
+		it("rio-standard user should be able to create privileges it already has 2", func() {
+			testService.Kubeconfig = standardUser.GetKubeconfig()
+			testService.Create(t, "--permission", "watch rio.cattle.io/app", "--disable-service-mesh", "nginx")
+		})
+
+	}, spec.Parallel(), spec.Flat())
```
cbron: Curious if there is a reason you're using spec.flat ? Thats the default.

tests/testutil/service.go
```diff
@@ -4,23 +4,54 @@ import (
...
+	if err != nil {
+		ts.T.Fatalf("Failed to create service:  %v", err.Error())
+	}
+	err = ts.waitForReadyService()
+	if err != nil {
+		ts.T.Fatalf(err.Error())
+	}
+}
+
+func (ts *TestService) CreateWithError(t *testing.T, source ...string) error {
```
cbron: Can you add a comment here just saying that this duplicates `Create` except that it returns an error.

tests/integration/rbac_test.go
StrongMonkey: Noticed that without this Flat the parent context(Before) will be executed on every run.

cbron: I believe the before block will run on every `it` block with either flat or nested. Afaik flat just means that all these tests aren't sub-scoped and can be run in parallel with any other suite's test.

PR #550, opened 2019-10-04
Automation scale and run tests

izaac:
Porting run tests from python to go spec.
Adding scale up to 3 test.
Removing python script.

Related to this I had to remove all the python test objects and disable their execution from CI because `py.test` was failing with error code 5 (no tests to run).


tests/integration/scale_test.go
```diff
@@ -34,7 +34,11 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
 			assert.Equal(t, 0, service.GetAvailableReplicas())
 			assert.Equal(t, 0, service.GetScale())
 		})
-		it.Pend("should scale to x,y,z....", func() {
+		it("should scale to 3", func() {
+			assert.Equal(t, 1, service.GetAvailableReplicas())
+			service.Scale(3)
+			assert.Equal(t, 3, service.GetAvailableReplicas())
+			assert.Equal(t, 3, service.GetScale())
```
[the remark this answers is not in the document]
izaac: Updated with the scenarios from `test_scale.py` and removed the python script too.

tests/integration/run_test.go
```diff
@@ -24,10 +24,10 @@ func runTests(t *testing.T, when spec.G, it spec.S) {
 	when("rio run is called", func() {
 		it("should create a service with default specifications", func() {
 			assert.Equal(t, 1, service.GetAvailableReplicas(), "should have one available replica")
-			// TODO: check the following and then remove test_run.py
-			// scale of 1
-			// weight of 100 in spec
-			// image == "nginx"
+			service.Scale(1)
```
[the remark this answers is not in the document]
izaac: I see that's a good point.

scripts/test
```diff
@@ -19,14 +19,4 @@ if [[ ${ARCH} == amd64 ]]; then
     rio install --check >/dev/null 2>&1
 
     go test -v ./tests/integration/... --integration-tests
-
```
cbron: Nice, thanks for removing this too.

tests/integration/scale_test.go
```diff
@@ -21,7 +21,7 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
 		service.Remove()
 	})
 
-	when("scale is called on a service", func() {
+	when.Focus("scale is called on a service", func() {
```
cbron: Remove `.Focus`

izaac: Good catch, removed.

tests/integration/scale_test.go
```diff
@@ -34,7 +34,23 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
...
+			assert.Equal(t, 1, service.GetAvailableReplicas())
+			service.Scale(5)
+			assert.Equal(t, 5, service.GetAvailableReplicas())
+			assert.Equal(t, 5, service.GetScale())
+		})
+		it("should scale to 10", func() {
+			assert.Equal(t, 1, service.GetAvailableReplicas())
+			service.Scale(10)
+			assert.Equal(t, 10, service.GetAvailableReplicas())
+			assert.Equal(t, 10, service.GetScale())
```
rancher-max: On my k3s cluster, tests for setting scale above 2 were a bit flaky. I had to update /tests/testutil/service.go in the `waitForScale` function to wait for 180 seconds at most (line 261)

cbron: Added fix for that here: https://github.com/rancher/rio/pull/552/files#diff-ebf1c8ea2e75e3983bbc0d00d1ed73e0R240

In short: it will wait `scale * 20 seconds`, or 2 minutes, whichever is greater. In this case it will wait 200 seconds. May need to bump that up to 25 or 30 seconds depending on systems we are targeting.

PR #552, opened 2019-10-07
Use k8s wait method and scale wait times to remove flakiness

cbron:
Follow up to https://github.com/rancher/rio/issues/486 to remove flakiness 

Reuse poll to make life easier: https://godoc.org/k8s.io/apimachinery/pkg/util/wait#Poll

### DIFF
--- a/tests/testutil/publicdomain.go
+++ b/tests/testutil/publicdomain.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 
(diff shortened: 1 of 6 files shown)


tests/testutil/publicdomain.go
```diff
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
+
```
daxmc99: Nit: Not sure if we have decided how we are grouping imports but I prefer them grouped std lib & other

cbron: Addressed in person, this is 3 leveled version with ours 3rd, I'll move to 2 level later on.

PR #553, opened 2019-10-07
Add wrapper for kubectl

rancher-max:
Beginning of validations for Issue #559 
This creates the kubectl wrapper and a couple of simple tests using it

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -2,7 +2,7 @@ FROM golang:1.12.6-alpine3.9
 
 ARG DAPPER_HOST_ARCH
 ENV ARCH $DAPPER_HOST_ARCH
-ENV K3S_VERSION v0.8.0
+ENV K3S_VERSION v0.9.1
 ENV KUBECTL_VERSION v1.15.0
 
 RUN apk -U add bash git gcc musl-dev docker vim less file curl wget ca-certificates python3 py3-pip python3-dev openssl-dev libffi-dev jq
@@ -21,7 +21,7 @@ RUN if [ "${ARCH}" == "amd64" ]; then \
     fi
 RUN curl -sL https://github.com/rancher/k3s/releases/download/${K3S_VERSION}/k3s > /usr/bin/k3s && \
     chmod +x /usr/bin/k3s
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl > /usr/bin/kubectl && \
+RUN curl -sL https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl > /usr/bin/kubectl && \
     chmod +x /usr/bin/kubectl
 
 ENV DAPPER_ENV REPO TAG DRONE_TAG CROSS
--- a/tests/integration/config_test.go
+++ b/tests/integration/config_test.go
@@ -22,6 +22,7 @@ func configTests(t *testing.T, when spec.G, it spec.S) {
 			testText := []string{"a=b", "foo=bar"}
 			config.Create(t, testText)
 			assert.Equal(t, testText, config.GetContent())
+			assert.Equal(t, testText, config.GetKubeContent())
 		})
 	}, spec.Parallel())
 }
--- a/tests/integration/run_test.go
+++ b/tests/integration/run_test.go
@@ -23,9 +23,12 @@ func runTests(t *testing.T, when spec.G, it spec.S) {
 
 	when("rio run is called", func() {
 		it("should create a service with default specifications", func() {
+			runningPods := service.GetRunningPods()
 			assert.Equal(t, 1, service.GetAvailableReplicas(), "should have one available replica")
 			assert.Equal(t, 100, service.GetCurrentWeight())
 			assert.Equal(t, "nginx", service.GetImage())
+			assert.Contains(t, runningPods, service.App.Name)
+			assert.Contains(t, runningPods, "2/2")
 		})
 	})
 }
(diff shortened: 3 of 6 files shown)


tests/integration/run_test.go
```diff
@@ -23,9 +23,12 @@ func runTests(t *testing.T, when spec.G, it spec.S) {
 
 	when("rio run is called", func() {
 		it("should create a service with default specifications", func() {
+			runningPods := service.GetRunningPods()
 			assert.Equal(t, 1, service.GetAvailableReplicas(), "should have one available replica")
 			assert.Equal(t, 100, service.GetCurrentWeight())
 			assert.Equal(t, "nginx", service.GetImage())
+			assert.Contains(t, runningPods, service.App.Name)
+			assert.Contains(t, runningPods, "2/2")
```
cbron: Was thinking there should be better way to test this, but looks like ready pods summary isn't available via json without calculating yourself.

Dockerfile.dapper
```diff
@@ -21,7 +21,7 @@ RUN if [ "${ARCH}" == "amd64" ]; then \
     fi
 RUN curl -sL https://github.com/rancher/k3s/releases/download/${K3S_VERSION}/k3s > /usr/bin/k3s && \
     chmod +x /usr/bin/k3s
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl > /usr/bin/kubectl && \
+RUN curl -sL https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl > /usr/bin/kubectl && \
```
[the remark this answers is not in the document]
rancher-max: This change was at the suggestion of @StrongMonkey to fix this issue with kubectl in CI:
`Failed to get pods: fork/exec /usr/bin/kubectl: exec format error:`

tests/testutil/config.go
```diff
@@ -23,14 +24,14 @@ type TestConfig struct {
 // This does not take a name or namespace param, that is setup by default
 func (tc *TestConfig) Create(t *testing.T, content []string) {
 	tc.T = t
-	name := GenerateName()
-	tc.Name = fmt.Sprintf("%s/%s", testingNamespace, name)
-	err := tc.createTempFile(name, content)
+	tc.Name = GenerateName()
+	tc.FullName = fmt.Sprintf("%s/%s", testingNamespace, tc.Name)
+	err := tc.createTempFile(tc.Name, content)
```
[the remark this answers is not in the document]
rancher-max: Agreed. I'll revert this and look into updating domain.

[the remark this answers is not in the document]
cbron: You can leave domain for now, thats minor issue.

tests/testutil/service.go
```diff
@@ -171,6 +171,21 @@ func (ts *TestService) GetScale() int {
 	return *ts.Service.Spec.Scale
 }
 
+func (ts *TestService) GetRunningPods() string {
+	ts.reloadApp()
```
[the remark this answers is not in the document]
rancher-max: I reload the app here because of the need for `ts.App.Name` call on line 178, and it's empty before calling `ts.reloadApp()`

tests/integration/run_test.go
rancher-max: Technically, we can validate this by getting the pods at `.items[*].status.containerStatuses` at either ready or state. I feel like in this case the 2/2 is a bit more clear and concise, but am open to changing it.

tests/integration/config_test.go
```diff
@@ -18,10 +18,11 @@ func configTests(t *testing.T, when spec.G, it spec.S) {
 	})
 
 	when("a config is created with data", func() {
-		it("should contain that data", func() {
+		it.Focus("should contain that data", func() {
```
cbron: Focus removal

tests/testutil/config.go
```diff
@@ -114,3 +126,18 @@ func (tc *TestConfig) waitForConfig() error {
 	}
 	return nil
 }
+
+// GetContent returns the configs Data.Content as list of strings, newline separated
+func (tc *TestConfig) getContentData(cm corev1.ConfigMap) []string {
```
cbron: Should probably keep this as standalone method, off the TestConfig struct, to avoid confusion.

tests/testutil/config.go
```diff
@@ -53,17 +53,29 @@ func (tc *TestConfig) Remove() {
...
+	if err != nil {
+		tc.T.Fatalf("Failed to unmarshal ConfigMaps result: %s with error: %v", resultString, err.Error())
+	}
+
+	return results
+}
+
+// GetKubeContent returns the kubectl configmap's Data.Content as list of strings, newline separated
+func (tc *TestConfig) GetKubeContent() []string {
+	return tc.getContentData(tc.KubectlGet())
```
cbron: Any reason we can't merge `KubectlGet` and `GetKubeContent` ?

rancher-max: We can. I was thinking the GetKubeContent would be useful for the future in case any tests get added that want to validate something other than content. Would it make more sense to add it at that time?

cbron: I would say yes, just to avoid premature optimization.

Dockerfile.dapper
```diff
@@ -2,7 +2,7 @@ FROM golang:1.12.6-alpine3.9
 
 ARG DAPPER_HOST_ARCH
 ENV ARCH $DAPPER_HOST_ARCH
-ENV K3S_VERSION v0.8.0
+ENV K3S_VERSION v0.9.1
```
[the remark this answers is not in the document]
cbron: He's good with it.

Issue #551, opened 2019-10-05
rio export --stack only captures one externalservice and one route grouping


rancher-max:
**Describe the bug**
When using `rio export --stack <namespace>`, the generated Riofile only has one externalservice and one route grouping, even if multiple of each of these has been created.

I executed this on a k3s cluster with 1 node.

**To Reproduce**
1. Build a few services, externalservices, and routes. To make it easy, use this command from this repo and branch: `rio up --name fullstack --branch k3s-branch https://github.com/rancher-max/riofile-demo`
2. Let it build, then run `rio export --stack default`
    - Result only has one externalservice and one route grouping.

**Expected behavior**
I expect the exact same Riofile I used to be generated. It should have the other route and other externalservice present.

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-19T13:57:45Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.4-k3s.1", GitCommit:"a7531b1ab3fd5ff987b074472ddfb84a2f5326bc", GitTreeState:"clean", BuildDate:"2019-09-19T22:36Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.5.0 (303f3652)
Rio CLI Version: v0.5.0 (303f3652)
Cluster Domain: x.on-rio.io
Cluster Domain IPs: x.x.x.x
System Namespace: rio-system

System Components:
activator status: running
autoscaler status: running
build-controller status: running
buildkitd-dev status: running
cert-manager status: running
controller status: running
gateway status: running
istio-pilot status: running
registry status: running
socat-socket status: running
webhook status: running
```
**Additional context**
`rio systemlogs`
output:
```
rio-controller | E1004 23:55:32.857163       1 controller.go:117] error syncing 'default/d-test': handler routing-serviceset-object-controller: Operation cannot be fulfilled on  "routing-serviceset": delaying object set, handler routing-serviceset-object-controller: Operation cannot be fulfilled on  "routing-serviceset": delaying object set, requeuing
[...]

PR #555, opened 2019-10-07
Fix bug to allow riofile export to have multiples

cbron:
Issue: https://github.com/rancher/rio/issues/551

We are adding the memory address of iterator `obj` twice, but then re-setting obj on loop so you get multiple of the same. Fix is to use address of actual value, not the iterator. 

Will add testing for this in the export roundtrip issues. 

Also add stack to rm types in CLI

### DIFF
--- a/cli/cmd/export/export.go
+++ b/cli/cmd/export/export.go
@@ -101,28 +101,28 @@ func collectObjectsFromNs(ctx *clicontext.CLIContext, ns string) ([]runtime.Obje
 		objects = append(objects, &services.Items[i])
 	}
 
-	for _, svc := range services.Items {
-		configs, err := configmapsFromService(ctx, svc)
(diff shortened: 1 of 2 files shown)


cli/cmd/export/export.go
```diff
@@ -101,28 +101,28 @@ func collectObjectsFromNs(ctx *clicontext.CLIContext, ns string) ([]runtime.Obje
 		objects = append(objects, &services.Items[i])
 	}
 
-	for _, svc := range services.Items {
-		configs, err := configmapsFromService(ctx, svc)
+	for i := range services.Items {
+		configs, err := configmapsFromService(ctx, services.Items[i])
```
cbron: This one isn't technically necessary since we aren't using it's pointer, but it future proofs us.

Issue #559, opened 2019-10-09
Add kubectl validations to the test framework


rancher-max:
We need to validate the objects that are created not only with rio but also with kubectl. The underlying kubernetes objects are very important in our test validations to ensure services are up and running as we would expect.

izaac:
Last set of validations are in closing @rancher-max

PR #560, opened 2019-10-10
Adding extra validations using kubectl for service endpoints

izaac:
This validates the endpoint URLs generated by rio for the revisions and not versioned are present.

https://github.com/rancher/rio/issues/559

### DIFF
--- a/tests/integration/endpoint_test.go
+++ b/tests/integration/endpoint_test.go
@@ -14,6 +14,9 @@ func endpointTests(t *testing.T, when spec.G, it spec.S) {
 	var service testutil.TestService
 	var stagedService testutil.TestService
 
+	it.Before(func() {
+		service.Create(t, "ibuildthecloud/demo:v1")
+	})
 	it.After(func() {
 		service.Remove()
 		stagedService.Remove()
@@ -21,10 +24,29 @@ func endpointTests(t *testing.T, when spec.G, it spec.S) {
 
 	when("a service is running and another is staged", func() {
 		it("should have endpoints that are available on both", func() {
-			service.Create(t, "ibuildthecloud/demo:v1")
 			stagedService = service.Stage("ibuildthecloud/demo:v3", "v3")
-			assert.Equal(t, "Hello World", service.GetEndpoint())
-			assert.Equal(t, "Hello World v3", stagedService.GetEndpoint())
+
+			// Check the hostnames returned by Rio and Kubectl are equal
+			assert.Equal(t,
+				testutil.GetHostname(service.GetKubeEndpointURL()),
+				testutil.GetHostname(service.GetEndpointURL()),
+			)
+			assert.Equal(t,
+				testutil.GetHostname(stagedService.GetKubeEndpointURL()),
+				testutil.GetHostname(stagedService.GetEndpointURL()),
+			)
+
+			assert.Equal(t, "Hello World", service.GetEndpointResponse())
+			assert.Equal(t, "Hello World v3", stagedService.GetEndpointResponse())
+		})
+		it("should have the service app endpoint properly created", func() {
+
+			// Check the hostnames returned by Rio and Kubectl are equal
+			assert.Equal(t,
+				testutil.GetHostname(service.GetKubeAppEndpointURL()),
+				testutil.GetHostname(service.GetAppEndpointURL()),
+			)
+			assert.Equal(t, "Hello World", service.GetAppEndpointResponse())
 		})
 	}, spec.Parallel())
 }
--- a/tests/testutil/service.go
+++ b/tests/testutil/service.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"strings"
 	"testing"
 	"time"
 
@@ -133,20 +134,6 @@ func (ts *TestService) Stage(source, version string) TestService {
 	return stagedService
 }
 
-// GetEndpoint performs an http.get against the service endpoint and returns response if
-// status code is 200, otherwise it errors out
-func (ts *TestService) GetEndpoint() string {
-	endpoint, err := ts.waitForEndpointDNS()
-	if err != nil {
-		ts.T.Fatal(err.Error())
-	}
-	response, err := WaitForURLResponse(endpoint)
-	if err != nil {
-		ts.T.Fatal(err.Error())
-	}
-	return response
-}
-
 // Export calls "rio export {serviceName}" and returns that in a new TestService object
 func (ts *TestService) Export() TestService {
 	args := []string{"export", "--type", "service", "--format", "json", ts.Name}
@@ -170,6 +157,24 @@ func (ts *TestService) ExportRaw() TestService {
 //////////
 // Getters
 //////////
+
+// GetEndpointResponse performs an http.get against the service endpoint and returns response if
+// status code is 200, otherwise it errors out
+func (ts *TestService) GetEndpointResponse() string {
+	response, err := WaitForURLResponse(ts.GetEndpointURL())
+	if err != nil {
+		ts.T.Fatal(err.Error())
+	}
+	return response
+}
+
+func (ts *TestService) GetAppEndpointResponse() string {
+	response, err := WaitForURLResponse(ts.GetAppEndpointURL())
+	if err != nil {
+		ts.T.Fatal(err.Error())
+	}
+	return response
+}
 
 // Returns count of ready and available pods
 func (ts *TestService) GetAvailableReplicas() int {
@@ -219,6 +224,68 @@ func (ts *TestService) GetRunningPods() string {
 		ts.T.Fatalf("Failed to get running pods:  %v", err.Error())
 	}
 	return out
+}
+
+// GetEndpointURL returns the URL for this service's app
+func (ts *TestService) GetEndpointURL() string {
+	url, err := ts.waitForEndpointDNS()
+	if err != nil {
+		ts.T.Fatalf("Failed to get the endpoint url:  %v", err.Error())
+		return ""
+	}
+	return url
+}
+
+// GetAppEndpointURL retrieves the service's app endpoint URL and returns it as string
+func (ts *TestService) GetAppEndpointURL() string {
+	url, err := ts.waitForAppEndpointDNS()
+	if err != nil {
+		ts.T.Fatalf("Failed to get the endpoint url:  %v", err.Error())
+		return ""
+	}
+	return url
+}
+
+// GetKubeEndpointURL returns the app revision endpoint URL
+// and returns it as string
+func (ts *TestService) GetKubeEndpointURL() string {
+	_, err := ts.waitForEndpointDNS()
+	if err != nil {
+		ts.T.Fatalf("Failed waiting for DNS:  %v", err.Error())
+		return ""
+	}
+	args := []string{"get", "service.rio.cattle.io",
+		"-n", testingNamespace,
+		ts.Service.Name,
+		"-o", `jsonpath="{.status.endpoints[0]}"`}
+	url, err := KubectlCmd(args)
+	if err != nil {
+		ts.T.Fatalf("Failed to get endpoint url:  %v", err.Error())
+		return ""
+	}
+	return strings.Replace(url, "\"", "", -1) // remove double quotes from output
+}
+
+// GetKubeAppEndpointURL returns the endpoint URL of the service's app
+// by using kubectl and returns it as string
+func (ts *TestService) GetKubeAppEndpointURL() string {
+	_, err := ts.waitForAppEndpointDNS()
+	if err != nil {
+		ts.T.Fatalf("Failed waiting for DNS:  %v", err.Error())
+		return ""
+	}
+	appName := strings.Split(ts.AppName, "/")[1]
+	args := []string{"get", "apps",
+		"-n", testingNamespace,
+		appName,
+		"-o", `jsonpath="{.status.endpoints[0]}"`}
+	url, err := KubectlCmd(args)
+	if err != nil {
+		ts.T.Fatalf("Failed to get app endpoint url:  %v", err.Error())
+		return ""
+	}
+
+	return strings.Replace(url, "\"", "", -1) // remove double quotes from output
 }
 
 //////////////////
@@ -371,4 +438,24 @@ func (ts *TestService) waitForEndpointDNS() (string, error) {
 		return "", errors.New("service endpoint never created")
 	}
 	return ts.Service.Status.Endpoints[0], nil
+}
+
+func (ts *TestService) waitForAppEndpointDNS() (string, error) {
+	if len(ts.App.Status.Endpoints) > 0 {
+		return ts.App.Status.Endpoints[0], nil
+	}
+	f := wait.ConditionFunc(func() (bool, error) {
+		err := ts.reloadApp()
+		if err == nil {
+			if len(ts.App.Status.Endpoints) > 0 {
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+	err := wait.Poll(2*time.Second, 60*time.Second, f)
+	if err != nil {
+		return "", errors.New("app endpoint never created")
+	}
+	return ts.App.Status.Endpoints[0], nil
 }
--- a/tests/testutil/testutil.go
+++ b/tests/testutil/testutil.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"math/rand"
 	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"strings"
@@ -101,4 +102,12 @@ func RandomString(length int) string {
 
 func GenerateName() string {
 	return strings.Replace(namesgenerator.GetRandomName(2), "_", "-", -1)
+}
+
+func GetHostname(URL string) string {
+	u, err := url.Parse(URL)
+	if err != nil {
+		return ""
+	}
+	return u.Hostname()
 }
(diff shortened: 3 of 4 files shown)


tests/integration/endpoint_test.go
```diff
@@ -14,17 +14,24 @@ func endpointTests(t *testing.T, when spec.G, it spec.S) {
...
 		service.Remove()
 		stagedService.Remove()
 	})
 
 	when("a service is running and another is staged", func() {
 		it("should have endpoints that are available on both", func() {
-			service.Create(t, "ibuildthecloud/demo:v1")
-			stagedService = service.Stage("ibuildthecloud/demo:v3", "v3")
-			assert.Equal(t, "Hello World", service.GetEndpoint())
+						assert.Equal(t, "Hello World", service.GetEndpoint())
```
cbron: Spacing here is off.

tests/integration/endpoint_test.go
```diff
@@ -14,17 +14,24 @@ func endpointTests(t *testing.T, when spec.G, it spec.S) {
...
 			assert.Equal(t, "Hello World v3", stagedService.GetEndpoint())
+			assert.Equal(t, service.GetKubeEndpointUrlByRevision("v0"), service.GetEndpointUrl())
```
cbron: Should we test against the wanted "hello world" string instead ? If both endpoints return nothing or a 500, are we ok with that as long as they are the same ? I know we do a check above that makes that question void but it's possible in the future it gets removed for some reason.

tests/testutil/service.go
```diff
@@ -202,6 +203,76 @@ func (ts *TestService) GetRunningPods() string {
...
+	return url
+}
+
+func (ts *TestService) GetKubeEndpointUrlByRevision(revision string) string {
+	err := ts.reloadApp()
+	if err != nil {
+		ts.T.Fatalf("Failed to reload app:  %v", err.Error())
+		return ""
+	}
+	args := append([]string{"--format", "json",
```
cbron: Don't need append here, just do `args := []string....`. There was some leftover code that did this by accident that is being removed.

tests/testutil/service.go
```diff
@@ -202,6 +203,76 @@ func (ts *TestService) GetRunningPods() string {
...
+		fmt.Sprintf("%s:%s", strings.Split(ts.Name, ":")[0], revision)})
+	out, err := RioCmd("inspect", args)
+	ts.Service = riov1.Service{}
```
cbron: For the kube methods, lets put them in variables and return or work with the variables, not on the ts object itself. So `s := riov1.Service{}` and then use `s`.

tests/testutil/service.go
```diff
@@ -202,6 +203,76 @@ func (ts *TestService) GetRunningPods() string {
...
+	args = append([]string{"get", "service.rio.cattle.io",
+		"-n", testingNamespace,
+		ts.Service.Name,
+		"-o", `jsonpath="{.status.endpoints[0]}"`})
+	url, err := KubectlCmd(args)
+	if err != nil {
+		ts.T.Fatalf("Failed to endpoint url:  %v", err.Error())
+		return ""
+	}
+	url = url[1 : len(url)-1] // remove double quotes from output
```
cbron: Can we use `strings.Replace(url, "\"", "", -1)` instead or will that mess with other quotes ? We should at least ensure that `url != ""` before doing this or it will panic.

tests/testutil/service.go
```diff
@@ -202,6 +203,76 @@ func (ts *TestService) GetRunningPods() string {
...
+	}
+	url, err := ts.waitForEndpointDNS()
+	if err != nil {
+		ts.T.Fatalf("Failed to get the endpoint url:  %v", err.Error())
+		return ""
+	}
+	return url
+}
+
+func (ts *TestService) GetKubeEndpointUrlByRevision(revision string) string {
```
cbron: Can you add comments to these new methods describing what they do and what they return

tests/testutil/service.go
```diff
@@ -202,6 +203,76 @@ func (ts *TestService) GetRunningPods() string {
...
+	url, err := KubectlCmd(args)
+	if err != nil {
+		ts.T.Fatalf("Failed to endpoint url:  %v", err.Error())
+		return ""
+	}
+	url = url[1 : len(url)-1] // remove double quotes from output
+	return url
+}
+
+func (ts *TestService) GetKubeGlobalServiceEndpoint() string {
```
cbron: Naming of "global" here and below feels off, maybe call it "app" or we can call the other one "versioned" and this one not versioned.

tests/testutil/testutil.go
```diff
@@ -55,6 +55,22 @@ func KubectlCmd(args []string) (string, error) {
 	return string(stdOutErr), nil
 }
 
+type waitForMe = func() bool
```
cbron: This has been updated yesterday, we are using k8's wait method now: https://github.com/rancher/rio/blob/master/tests/testutil/externalservice.go#L79-L90

So you can remove this.

tests/testutil/service.go
```diff
@@ -202,6 +203,76 @@ func (ts *TestService) GetRunningPods() string {
...
+		return ""
+	}
+	args := append([]string{"--format", "json",
+		fmt.Sprintf("%s:%s", strings.Split(ts.Name, ":")[0], revision)})
+	out, err := RioCmd("inspect", args)
+	ts.Service = riov1.Service{}
+	if err := json.Unmarshal([]byte(out), &ts.Service); err != nil {
+		return ""
+	}
+	args = append([]string{"get", "service.rio.cattle.io",
```
cbron: If we reuse this over and over again, we should put it in a const by `testingNamespace`, but we can leave for now.

tests/integration/endpoint_test.go
izaac: @cbron I think I've addressed your feedback in a new commit.

tests/integration/endpoint_test.go
```diff
@@ -14,17 +14,33 @@ func endpointTests(t *testing.T, when spec.G, it spec.S) {
...
 			assert.Equal(t, "Hello World", service.GetEndpoint())
 			assert.Equal(t, "Hello World v3", stagedService.GetEndpoint())
+			kubeUrlv0 := service.GetKubeEndpointUrlByRevision("v0")
+			kubeUrlv3 := stagedService.GetKubeEndpointUrlByRevision("v3")
+			rioUrlv0 := service.GetEndpointUrl()
+			rioUrlv3 := stagedService.GetEndpointUrl()
+			assert.Equal(t, kubeUrlv0, rioUrlv0)
+			assert.Equal(t, kubeUrlv3, rioUrlv3)
+
+			assert.Equal(t, service.GetEndpointResponse(kubeUrlv0), service.GetEndpointResponse(rioUrlv0))
```
cbron: This whole block and associated methods aren't intuitive anymore, and there is some overlap in functionality across methods. What if we do: 
```
testUtil.WaitForURLResponse(service.GetEndpointURL) == "Hello World"
testUtil.WaitForURLResponse(stagedService.GetEndpointURL) == "Hello World V3"
service.GetEndpointURL == service.GetKubeEndpointURL
stagedService.GetEndpointURL == stagedService.GetKubeEndpointURL
```
Then for global test below lets do same: 
```
testUtil.WaitForURLResponse(service.GetAppEndpointURL) == "Hello World"
service.GetAppEndpointURL == service.GetKubeAppEndpointURL
```
That way we only have 2 real methods: endpointURL and appEndpointURL, and the equivalent kube versions of them. Might be contradicting myself, but we don't need to do a check on KubeEndpoint response if the URLs are equivalent.

tests/integration/endpoint_test.go
```diff
@@ -14,17 +14,33 @@ func endpointTests(t *testing.T, when spec.G, it spec.S) {
...
+		})
+		it("should have the global endpoint properly created", func(){
```
cbron: Lets move stagedService above so it doesn't get created here, only in first test. Might break this one actually if it does, as weight can move over to staged one and change endpoint response.

tests/testutil/service.go
```diff
@@ -121,7 +122,13 @@ func (ts *TestService) GetEndpoint() string {
 	if err != nil {
 		ts.T.Fatal(err.Error())
 	}
-	response, err := WaitForURLResponse(endpoint)
+	return ts.GetEndpointResponse(endpoint)
+}
+
+// GetEndpointResponse performs an http.get against the service endpoint and returns response if
```
cbron: Can you move all the "getEndpoint..." methods to the getters section below this so they are all in same location.

tests/testutil/service.go
```diff
@@ -202,6 +209,88 @@ func (ts *TestService) GetRunningPods() string {
...
+	if err != nil {
+		ts.T.Fatalf("Failed to get the endpoint url:  %v", err.Error())
+		return ""
+	}
+	return url
+}
+
+// GetKubeEndpointUrlByRevision returns the versioned endpoint URL by using the revision provided
+// and returns it as string
+func (ts *TestService) GetKubeEndpointUrlByRevision(revision string) string {
```
cbron: Don't need revision here, that lives on ts.Name. See above comment as this method will need to change anyway.

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
 	return out
 }
 
+// GetEndpointURL returns the URL for this service's app if the App object is versioned it returns
+// the versioned url.
```
cbron: This is incorrect, it only returns the service url. Service == versioned, app == unversioned. Let's not use the 'versioned' or 'unversioned' phrases, that can get confusing, let's just use 'service' or 'app'.

tests/testutil/service.go
```diff
@@ -152,6 +139,20 @@ func (ts *TestService) ExportRaw() TestService {
 // Getters
 //////////
 
+// GetEndpoint performs an http.get against the service endpoint and returns response if
+// status code is 200, otherwise it errors out
+func (ts *TestService) GetEndpoint() string {
+	endpoint, err := ts.waitForEndpointDNS()
```
cbron: I like how below in `GetAppEndpointResponse` you reuse the `GetAppEndpoint` method to get the url. Lets follow that pattern here and use `GetEndpointURL` to get the url and pass that into the Wait method.

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
...
+
+// GetAppEndpoint retrieves the not versioned endpoint URL of the service's app
+// by using rio's App object and returns it as string
+func (ts *TestService) GetAppEndpoint() string {
+	err := ts.reloadApp()
+	if err != nil {
+		ts.T.Fatalf("Failed to reload app:  %v", err.Error())
+		return ""
+	}
+	return ts.App.Status.Endpoints[0]
```
cbron: This is good, however it may error out if DNS isn't ready yet. Follow the pattern in `waitForEndpointDNS` and make a `waitForAppEndpointDNS`

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
...
+		return ""
+	}
+	if url != "" {
+		url = strings.Replace(url, "\"", "", -1) // remove double quotes from output
+	}
+	return url
+}
+
+// GetAppEndpoint retrieves the not versioned endpoint URL of the service's app
+// by using rio's App object and returns it as string
```
cbron: > // GetAppEndpoint retrieves the service's app endpoint URL and returns it as string

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
...
+		ts.T.Fatalf("Failed to get the endpoint url:  %v", err.Error())
+		return ""
+	}
+	return url
+}
+
+// GetKubeEndpointUrlByRevision returns the versioned endpoint URL by using the revision provided
+// and returns it as string
+func (ts *TestService) GetKubeEndpointURL() string {
+	err := ts.reloadApp()
```
cbron: Believe you'll have to have both of these kube methods use wait helpers also, since DNS can take a while to come up.

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
...
+	if err != nil {
+		ts.T.Fatalf("Failed to reload app:  %v", err.Error())
+		return ""
+	}
+	args := []string{"--format", "json",
+		fmt.Sprintf("%s:%s",
+			strings.Split(ts.Name, ":")[0],
+			strings.Split(ts.Name, ":")[1])}
```
cbron: Confused at what this Sprintf does, it splits a string on colon and then re-adds it back with colon ?

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
...
+	out, err := RioCmd("inspect", args)
+	s := riov1.Service{}
+	if err := json.Unmarshal([]byte(out), &s); err != nil {
+		return ""
+	}
+	args = []string{"get", "service.rio.cattle.io",
+		"-n", testingNamespace,
+		s.Name,
+		"-o", `jsonpath="{.status.endpoints[0]}"`}
+	url, err := KubectlCmd(args)
```
cbron: Whats the point of the first Riocmd call and unmarshall ? Can't you just delete that entirely and use ts.Name ?

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
...
+	if err != nil {
+		ts.T.Fatalf("Failed to get endpoint url:  %v", err.Error())
+		return ""
+	}
+	if url != "" {
```
cbron: You can actually remove this because `strings.Replace` works on an empty string. It's just the slice operations that don't.

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
...
+
+// GetKubeAppServiceEndpoint returns the not versioned endpoint URL of the service's app
+// by using kubectl and returns it as string
+func (ts *TestService) GetKubeAppServiceEndpoint() string {
+	err := ts.reloadApp()
+	if err != nil {
+		ts.T.Fatalf("Failed to reload app:  %v", err.Error())
+		return ""
+	}
+	nameNoRevision := strings.Split(ts.Name, ":")[0]
```
cbron: Just use `ts.AppName`

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
...
+	if url != "" {
+		url = strings.Replace(url, "\"", "", -1) // remove double quotes from output
+	}
+	return url
+}
+
+// GetKubeAppServiceEndpoint returns the not versioned endpoint URL of the service's app
+// by using kubectl and returns it as string
+func (ts *TestService) GetKubeAppServiceEndpoint() string {
+	err := ts.reloadApp()
```
cbron: Same here, needs wait method.

tests/testutil/service.go
```diff
@@ -202,6 +203,98 @@ func (ts *TestService) GetRunningPods() string {
...
+	if err != nil {
+		ts.T.Fatalf("Failed to reload app:  %v", err.Error())
+		return ""
+	}
+	nameNoRevision := strings.Split(ts.Name, ":")[0]
+	serviceName := strings.Split(nameNoRevision, "/")[1]
+	args := []string{"get", "apps",
+		"-n", testingNamespace,
+		fmt.Sprint(serviceName),
```
cbron: Don't need fmt here, just pass `serviceName`

tests/testutil/service.go
izaac: I'm getting the actual name of the revision with rio, then I feed kubectl to get the endpoint. @StrongMonkey's suggestion.

tests/testutil/service.go
izaac: It made more sense back when the method had the revision parameter.

tests/testutil/service.go
cbron: You should be able to just use `ts.name` or `ts.version`, really anything you get from the unmarshall already exists on the `ts` object.

tests/testutil/service.go
```diff
@@ -202,6 +207,82 @@ func (ts *TestService) GetRunningPods() string {
...
+
+// GetKubeAppEndpointURL returns the endpoint URL of the service's app
+// by using kubectl and returns it as string
+func (ts *TestService) GetKubeAppEndpointURL() string {
+	_, err := ts.waitForAppEndpointDNS()
+	if err != nil {
+		ts.T.Fatalf("Failed waiting for DNS:  %v", err.Error())
+		return ""
+	}
+	serviceName := strings.Split(ts.AppName, "/")[1]
```
cbron: this var should be `appName`

tests/testutil/service.go
```diff
@@ -202,6 +207,82 @@ func (ts *TestService) GetRunningPods() string {
...
+	args := []string{"get", "apps",
+		"-n", testingNamespace,
+		serviceName,
+		"-o", `jsonpath="{.status.endpoints[0]}"`}
+	url, err := KubectlCmd(args)
+	if err != nil {
+		ts.T.Fatalf("Failed to get app endpoint url:  %v", err.Error())
+		return ""
+	}
+	if url != "" {
```
cbron: You can remove this check and just return the `strings.Replace` response.

tests/testutil/service.go
```diff
@@ -202,6 +207,82 @@ func (ts *TestService) GetRunningPods() string {
...
+	args = []string{"get", "service.rio.cattle.io",
+		"-n", testingNamespace,
+		s.Name,
+		"-o", `jsonpath="{.status.endpoints[0]}"`}
+	url, err := KubectlCmd(args)
+	if err != nil {
+		ts.T.Fatalf("Failed to get endpoint url:  %v", err.Error())
+		return ""
+	}
+	url = strings.Replace(url, "\"", "", -1) // remove double quotes from output
```
cbron: You can return this directly.

tests/testutil/service.go
izaac: using `ts.Service.Name`

PR #562, opened 2019-10-11
Add route test and kubectl validations

rancher-max:
Issue: #559 

Summary of PR:
Add another useful test for routes
Add kubectl validations for domain, weight, and route tests

### DIFF
--- a/tests/testutil/router.go
+++ b/tests/testutil/router.go
@@ -21,13 +21,16 @@ type TestRoute struct {
 
 // Executes "rio route add routename.testing-namespace/{routePath} to {service}"
 // This does not take a domain param, that is setup by default.
+// domain is optional, if empty it will generate a random domain.
 // routePath is optional, if empty it will set only domain.
-func (tr *TestRoute) Add(t *testing.T, routePath string, action string, target TestService) {
+func (tr *TestRoute) Add(t *testing.T, domain string, routePath string, action string, target TestService) {
 	tr.T = t
-	fakeDomain := RandomString(5)
+	if domain == "" {
+		domain = RandomString(5)
+	}
 	tr.Path = routePath
-	tr.Name = fmt.Sprintf("%s/%s", testingNamespace, fakeDomain)
-	route := fmt.Sprintf("%s.%s%s", fakeDomain, testingNamespace, routePath)
+	tr.Name = fmt.Sprintf("%s/%s", testingNamespace, domain)
+	route := fmt.Sprintf("%s.%s%s", domain, testingNamespace, routePath)
 	_, err := RioCmd([]string{"route", "add", route, action, target.Name})
 	if err != nil {
 		tr.T.Fatalf("route add command failed:  %v", err.Error())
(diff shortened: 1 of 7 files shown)


tests/testutil/router.go
```diff
@@ -21,10 +21,14 @@ type TestRoute struct {
 
 // Executes "rio route add routename.testing-namespace/{routePath} to {service}"
 // This does not take a domain param, that is setup by default.
+// domain is optional, if empty it will generate a random domain.
 // routePath is optional, if empty it will set only domain.
-func (tr *TestRoute) Add(t *testing.T, routePath string, action string, target TestService) {
+func (tr *TestRoute) Add(t *testing.T, domain string, routePath string, action string, target TestService) {
 	tr.T = t
-	fakeDomain := RandomString(5)
+	fakeDomain := domain
```
cbron: `fakeDomain` doesn't make sense anymore, just check `domain` and if empty generate into it.

tests/testutil/router.go
```diff
@@ -62,6 +66,40 @@ func (tr *TestRoute) GetEndpoint() string {
 	return response
 }
 
+// GetKubeEndpoints performs an http.get against the route's full domain and all paths on it
+// Returns responses if status code is 200 for all of them, otherwise it errors out
+func (tr *TestRoute) GetKubeEndpoints() []string {
```
cbron: As part of changes in different PR, rename this to `GetKubeEndpointResponse`, and `GetEndpoint` to `GetEndpointResponse`. Need to differentiate from methods that return just a URL.

tests/testutil/router.go
```diff
@@ -62,6 +66,40 @@ func (tr *TestRoute) GetEndpoint() string {
...
+		tr.T.Fatal("router has no endpoint")
+	}
+
+	// Store response from URLs of each path on the route
+	for _, path := range results.Spec.Routes {
+		res, err := WaitForURLResponse(fmt.Sprintf("%s%s", results.Status.Endpoints[0], path.Matches[0].Path.Exact))
+		if err != nil {
+			tr.T.Fatal(err.Error())
+		}
+		responses = append(responses, res)
```
cbron: I wonder if we should make this method single route based to match `GetEndpoint`, you just range through results till you get your wanted path. This is part of larger discussion around the tricky design of TestRouter and multiple paths.

tests/testutil/service.go
```diff
@@ -202,6 +199,23 @@ func (ts *TestService) GetRunningPods() string {
 	return out
 }
 
+// GetKubeCurrentWeight takes in a revision value and retrieves the actual current weight, not the spec (end-goal) weight
+func (ts *TestService) GetKubeCurrentWeight(revision string) int {
```
cbron: You are already on the service object so you don't need to pass in revision, use `ts.Version`.

rancher-max: I might need to rewrite this whole thing, or create an `app.go` helper file. This really acts on the "apps.rio.cattle.io" resource. 
For example, when called in one of the tests, it's called from the same service object:
```
assert.Equal(t, 95, service.GetKubeCurrentWeight(service.Version))
assert.Equal(t, 5, service.GetKubeCurrentWeight(stagedService.Version))
```
I'm kind of leaning towards a design of something like `service.GetApp()` and then have a `TestApp` object to call something like this on. Thoughts?

cbron: Shouldn't go app route because most likely app won't exist soon, it will just all be services linked together. 
Why can't you call `stagedService.GetKubeCurrentWeight()` ? service and stagedService are on same app.

rancher-max: ... because that's too simple and I overthought it 👍

tests/testutil/router.go
rancher-max: Yeah we can handle this the same way as the other change. Keep it simple. 2 routes, they'll both have the same domain, just do what `GetEndpoint(Response)` does with tr.Path and call `KubeGetEndpoint(Response)` on each separate route

Issue #532, opened 2019-09-27
Test Riofile Parsing


cbron:
The Riofile will have special syntax for things like routing. We should have automated tests to ensure we can parse those variants as well as the rest of the manifest.

davidnuzik:
@cbron to revisit after the incoming rc and reassess.

daxmc99:
This riofile causes a panic 
```yaml
services:
    emoji-svc:
        app: emoji-svc
        version: v0
        image: buoyantio/emojivoto-emoji-svc:v8
        ports:
        - 8080:8080/grpc,grpc
        env:
        - GRPC_PORT=8080
        
    voting-svc:
        app: voting-svc
        version: v0
        image: buoyantio/emojivoto-voting-svc:v8
        ports:
        - 8080:8080/grpc,grpc
        env:
        - GRPC_PORT=8080
        
    web-svc:
        app: web-svc
        version: v0
        image: buoyantio/emojivoto-web:v8
        ports:
            - 80:8080/http,web
        env:
        - WEB_PORT=8080
        - EMOJISVC_HOST=emoji-svc:8080
        - VOTINGSVC_HOST=voting-svc:8080
        - INDEX_BUNDLE=dist/index_bundle.js 
        
    vote-bot:
        app: vote-bot
        version: v0
        image: buoyantio/emojivoto-web:v8
        command:
        - emojivoto-vote-bot
        env:
        - WEB_HOST=web-svc:80

configs:
    config-foo:
    key1: |-
        I'm providing some data to your rio apps

routers:
    rio-router:
        to:
        - service: web-svc
          revision: v0
          namespace: default

```
stack trace 
```
default:stack/rio-demo
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/rancher/norman/pkg/types/mapper.SetValue.ToInternal(...)
        /Users/dax/go/pkg/mod/github.com/rancher/norman@v0.0.0-20191030191625-ebecbda5fbe3/pkg/types/mapper/set_value.go:22
github.com/rancher/norman/pkg/types.Mappers.ToInternal(0xc0003b7f40, 0xa, 0xa, 0x0, 0xc000542010, 0x0)
        /Users/dax/go/pkg/mod/github.com/rancher/norman@v0.0.0-20191030191625-ebecbda5fbe3/pkg/types/mapper.go:40 +0x8f
[...]

PR #564, opened 2019-10-11
WIP: add fuzz testing to test parser input

cbron:
Issue: https://github.com/rancher/rio/issues/532

Needs this, then a real vendor update: https://github.com/rancher/mapper/pull/3

Primarily this adds a framework for fuzzy testing and unit testing around that
It also  removes python from dapperfile.


cbron:
Will reopen in a new PR.

PR #568, opened 2019-10-14
Adding validations to scale tests

izaac:
This is part of #559 Add kubectl validations to the test framework

### DIFF
--- a/tests/integration/scale_test.go
+++ b/tests/integration/scale_test.go
@@ -14,7 +14,7 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
 	var service testutil.TestService
 
 	it.Before(func() {
-		service.Create(t, "nginx")
+		service.Create(t, "sangeetha/mytestcontainer")
 	})
 
 	it.After(func() {
@@ -26,6 +26,8 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
 			assert.Equal(t, 1, service.GetAvailableReplicas())
 			service.Scale(2)
 			assert.Equal(t, 2, service.GetAvailableReplicas())
+			assert.Equal(t, service.GetKubeAvailableReplicas(), service.GetAvailableReplicas())
+			assert.True(t, service.PodsResponsesMatchAvailableReplicas("/name.html", service.GetAvailableReplicas()))
 		})
 		// This is an important test because zero scale is wide ranging feature
 		it("should scale to zero", func() {
@@ -39,18 +41,24 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
 			service.Scale(3)
 			assert.Equal(t, 3, service.GetAvailableReplicas())
 			assert.Equal(t, 3, service.GetScale())
+			assert.Equal(t, service.GetKubeAvailableReplicas(), service.GetAvailableReplicas())
+			assert.True(t, service.PodsResponsesMatchAvailableReplicas("/name.html", service.GetAvailableReplicas()))
 		})
 		it("should scale to 5", func() {
 			assert.Equal(t, 1, service.GetAvailableReplicas())
 			service.Scale(5)
 			assert.Equal(t, 5, service.GetAvailableReplicas())
 			assert.Equal(t, 5, service.GetScale())
+			assert.Equal(t, service.GetKubeAvailableReplicas(), service.GetAvailableReplicas())
+			assert.True(t, service.PodsResponsesMatchAvailableReplicas("/name.html", service.GetAvailableReplicas()))
 		})
 		it("should scale to 10", func() {
 			assert.Equal(t, 1, service.GetAvailableReplicas())
 			service.Scale(10)
 			assert.Equal(t, 10, service.GetAvailableReplicas())
 			assert.Equal(t, 10, service.GetScale())
+			assert.Equal(t, service.GetKubeAvailableReplicas(), service.GetAvailableReplicas())
+			assert.True(t, service.PodsResponsesMatchAvailableReplicas("/name.html", service.GetAvailableReplicas()))
 		})
 	}, spec.Parallel())
 }
--- a/tests/testutil/service.go
+++ b/tests/testutil/service.go
@@ -9,9 +9,9 @@ import (
 	"testing"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/wait"
-
 	riov1 "github.com/rancher/rio/pkg/apis/rio.cattle.io/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 type TestService struct {
@@ -80,10 +80,7 @@ func (ts *TestService) Remove() {
 
 // Call "rio scale ns/service={scaleTo}"
 func (ts *TestService) Scale(scaleTo int) {
-	_, err := RioCmd([]string{
-		"scale",
-		fmt.Sprintf("%s=%d", ts.Name, scaleTo),
-	})
+	_, err := RioCmd([]string{"scale", fmt.Sprintf("%s=%d", ts.Name, scaleTo)})
 	if err != nil {
 		ts.T.Fatalf("scale command failed:  %v", err.Error())
 	}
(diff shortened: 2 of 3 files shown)


tests/integration/scale_test.go
```diff
@@ -14,7 +14,7 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
 	var service testutil.TestService
 
 	it.Before(func() {
-		service.Create(t, "nginx")
+		service.Create(t, "sangeetha/mytestcontainer")
```
cbron: We should avoid using personal containers in our tests, if someone leaves and deletes their stuff all our tests break. Let's see if we can add image rancher account, or set it up as a repo in rancher account and create it via build.

tests/testutil/service.go
```diff
@@ -263,6 +264,30 @@ func (ts *TestService) GetKubeEndpointURL() string {
 	return strings.Replace(url, "\"", "", -1) // remove double quotes from output
 }
 
+// GetKubeAvailableReplicas returns the app number of ready replicasets with kubectl
+// and returns it as integer
+func (ts *TestService) GetKubeAvailableReplicas() int {
```
cbron: This whole approach still feels odd, we should chat during standup, I'm still not convinced we need duplicated kube testing. If we do continue this route, instead of using kubectl can we use the kube clientset ? 
https://github.com/kubernetes/kubernetes/blob/master/test/integration/deployment/util.go#L61
https://github.com/kubernetes/kubernetes/blob/master/test/integration/deployment/util.go#L447-L453

tests/testutil/service.go
```diff
@@ -302,6 +327,26 @@ func (ts *TestService) GetKubeCurrentWeight() int {
 	return getRevisionWeight(app, ts.Version)
 }
 
+// PodsResponsesMatchAvailableReplicas does a GetURL in the App endpoint and stores the response in a slice
+// the length of the resulting slice should represent the number of responsive pods in a service.
+// Returns true if the number of replicas is equal to the length of the responses slice.
+func (ts *TestService) PodsResponsesMatchAvailableReplicas(path string, numberOfReplicas int) bool {
```
cbron: I don't understand this method, you'r hitting the app endpoint for the amount of ready replicas ? That could hit the same replica every time. Why not just hit each replica ? Whats the goal of this test ? We already test each service endpoint URL.

tests/integration/scale_test.go
izaac: Agree, I'm using it here because it's the same image used for validations in rancher/rancher

tests/integration/scale_test.go
```diff
@@ -26,6 +26,8 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
 			assert.Equal(t, 1, service.GetAvailableReplicas())
 			service.Scale(2)
 			assert.Equal(t, 2, service.GetAvailableReplicas())
+			assert.Equal(t, service.GetAvailableReplicas(), service.GetKubeAvailableReplicas())
+			assert.True(t, service.PodsResponsesMatchAvailableReplicas("/name.html", service.GetAvailableReplicas()))
```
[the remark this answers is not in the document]
izaac: I'm using a path because the image is using that html code to validate the pods hostnames.

tests/testutil/service.go
izaac: Making sure the number of replicas match the different hostname responses from the `name.html` from each.

tests/integration/scale_test.go
cbron: ok

tests/testutil/service.go
izaac: Uhm I understand your suggestion @cbron , use something like this https://github.com/kubernetes/client-go instead of invoking kubectl from command line? we should meet with @sangeethah about this, I'm open to follow that path if everyone is ok with it.

tests/testutil/service.go
cbron: If you're hitting app endpoint though, you could get the same replica every time. Even if you did split it out to hit different endpoints, only thing you're guaranteeing is that all pods are running.

izaac: I think the app endpoint is balancing the requests properly I did some manual testing with curl in a service with scale=5 and I was getting the 5 different hostnames. Sometimes of course duplicated but eventually a did get them.

That's why I'm multiplying the scale by 5, to have room for probability of getting all them.

cbron: You should be able to just get the individual pod endpoints and hit those. But even then it feels overkill to me, since this is all handled inside Kubernetes.

tests/testutil/service.go
cbron: That would likely be better

tests/testutil/service.go
izaac: But for validation we're also testing the balancer, this will be specially relevant for promotions: https://github.com/rancher/rio/issues/485

cbron: Isn't that just balancing between revisions though ? Not the pods inside the revisions.

izaac: I think that hits the latest promoted revision.

tests/testutil/service.go
```diff
@@ -302,6 +311,46 @@ func (ts *TestService) GetKubeCurrentWeight() int {
...
+}
+
+// KubeCompareReplicasValues get the app number of ready replicasets with a clientset
+// and returns true if that value match the scale given
+func (ts *TestService) KubeCompareReplicasValues(scale int) bool {
+	err := ts.waitForAvailableReplicas(scale)
+	if err != nil {
+		ts.T.Fatal(err.Error())
+	}
+	clientset := GetClientSet()
```
izaac: @cbron I've used a clientset here to get the replicas from k8s and compare it with the replica count reported by rio. 
For future test changes we might need to take some time and refactor the kubectl helper methods into their clientset alternatives. 
That would require time though (mostly figure out the proper clientset calls) and we can incrementally work on that on future PRs. Or work on a large change and refactor all of them at once.

rancher-max: I like using the clientset. I think in a future PR, we should update our current kubectl commands to use it.
For now, can we update the function name `GetClientSet()` to be more clear what the client set is: `GetKubeClient()` ? or something like that.

Also for consistency, I'd recommend changing this to return the available replicas, like the counterpart of `GetAvailableReplicas()`. This would be similar to how `GetAppEndpointURL()` and `GetKubeAppEndpointURL()` are.

tests/integration/scale_test.go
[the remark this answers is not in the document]
izaac: I think I'm staying with sangeetha's image, I don't see other in the rancher test framework and is also a very well known image for QA purposes within Rancher.

tests/testutil/service.go
izaac: I manually tested my previous comment and the App URL hits only the promoted revision when in 100% weight which is the case for these scenarios and only gets those app pods in the replicas. I think we would need separate requests methods when we deal with the weight/revision test cases.

tests/testutil/router.go
```diff
@@ -31,7 +33,11 @@ func (tr *TestRoute) Add(t *testing.T, domain string, routePath string, action s
 	tr.Path = routePath
 	tr.Name = fmt.Sprintf("%s/%s", testingNamespace, domain)
 	route := fmt.Sprintf("%s.%s%s", domain, testingNamespace, routePath)
-	_, err := RioCmd([]string{"route", "add", route, action, target.Name})
+	opts := wait.Backoff{Factor: 1.0, Steps: 3}
```
cbron: Does it make sense to just add this to RioCmd ?

tests/testutil/service.go
cbron: Ok. I still don't really like this approach, we can try it for now but if it gets flakey we'll need to redo. Entirely possible that if there is 5 replicas, 2 or 3 of them never get hit.

tests/testutil/service.go
```diff
@@ -302,6 +311,42 @@ func (ts *TestService) GetKubeCurrentWeight() int {
...
+// and returns true if that value match the scale given
+func (ts *TestService) GetKubeAvailableReplicas() int {
+	clientset := GetKubeClient()
+	replicaSetList, err := clientset.AppsV1().
+		ReplicaSets(testingNamespace).
+		List(metav1.ListOptions{LabelSelector: "app=" + ts.Service.Name})
+	if err != nil {
+		ts.T.Fatalf(err.Error())
+	}
+	if !(len(replicaSetList.Items) > 0) {
```
cbron: Does `if len == 0` not work here ? Don't think a negative is possible.

tests/testutil/testutil.go
```diff
@@ -111,3 +113,22 @@ func GetHostname(URL string) string {
 	}
 	return u.Hostname()
 }
+
+// stringInSlice returns true if string a value is equals to any element of the slice otherwise false
+func stringInSlice(a string, list []string) bool {
+	for _, b := range list {
```
cbron: nitpick, but can we rename `b` to `i` or `l`, that way it's somewhat relevant inside the loop.

tests/testutil/router.go
izaac: Actually @cbron I was discussing this one with @rancher-max but I'm kind of looking for an opinion on how to handle this one in the `RioCmd` method because it makes sense. 

But `cmd.CombinedOutput()` returns the stdout and we use that to print errors which we think are important for debugging.

```go
	var stdOutErr []byte
	opts := wait.Backoff{Factor: 1.0, Steps: 3}
	err := retry.RetryOnConflict(opts, func() error {
		stdOutErr, err := cmd.CombinedOutput()
		return err
	})
	if err != nil {
		return "", fmt.Errorf("%s: %s", err.Error(), stdOutErr)
	}
```
So how do we keep `stdOutError` in that case because `RetryOnConflict` only returns the errors after the retries.

tests/testutil/service.go
izaac: That's a valid point but in theory if we have enough requests we should hit all the pods because the balancer works in Round Robin  (unless keep-alive by the browser or session caching(?) from kubernetes)

This has been working fine locally and in CI but you're right there's a chance we don't hit all the pods. I could work on a better solution if that becomes a problem.

tests/testutil/router.go
cbron: Hmm true, `RetryOnConflict` func has only an error as return param. You could redo that logic (and `OnError`) in our framework, even though its unfortunate we can't reuse code.

izaac: In the meantime these were the most "conflicting" areas I identified on new clusters when running the tests in parallel. I can work in a reimplementation and refactor all these in a separate PR. Would you agree?

cbron: Sure. We can chat next week about approach too, see if there is a better way we aren't thinking of.

izaac: @rancher-max @cbron Ive added a simple function for retry in testutil. Seems to do the trick.

Issue #570, opened 2019-10-15
Automate p1 test scenarios 


rancher-max:
Automate the p1 test scenarios so that we can efficiently test the critical features of rio when any changes are made. We can eventually point the automation to different kubernetes cluster configurations to make rio more robust. 

When this is complete, the manual testing efforts will be able to be more focused on edge cases and less common scenarios, but we can have assurance that the main features of rio are intact.

rancher-max:
Completed and merged into master. Remaining P1s aren't as valid to automate in our integration tests.

PR #571, opened 2019-10-15
Add P1 tests to integration and validation

rancher-max:
Issue: #570 

This covers the first scenario from the P1 Rio Run Test Cases.
Some changes had to be made to the test framework to accommodate this. The three I'd like to highlight are:
- Installing the "hey" tool into CI to generate load on the endpoitns
- Removing the port from `service.Create()` and adding optional arguments to it. The port should be used as an argument in many cases, but not all. There are also additional run options that are important to test, so having the optional arguments will give us this ability.
-  Added this check in `waitForReadyService()`: `ts.Service.Status.DeploymentStatus.AvailableReplicas == ts.Service.Status.DeploymentStatus.Replicas`. This one I'm not convinced about, but want to get others' feedback. I think it might be better to have a `waitForReadyReplicas()` function that we call in `GetAvailableReplicas()` that has this logic in it. What do we all think?

### DIFF
--- a/tests/integration/domain_test.go
+++ b/tests/integration/domain_test.go
@@ -15,7 +15,7 @@ func domainTests(t *testing.T, when spec.G, it spec.S) {
 	var domain testutil.TestDomain
 
 	it.Before(func() {
-		service.Create(t, "nginx")
+		service.Create(t)
 	})
 
 	it.After(func() {
--- a/tests/integration/endpoint_test.go
+++ b/tests/integration/endpoint_test.go
@@ -14,16 +14,17 @@ func endpointTests(t *testing.T, when spec.G, it spec.S) {
 	var service testutil.TestService
 	var stagedService testutil.TestService
 
-	it.Before(func() {
-		service.Create(t, "ibuildthecloud/demo:v1")
-	})
-	it.After(func() {
-		service.Remove()
-		stagedService.Remove()
-	})
-
 	when("a service is running and another is staged", func() {
-		it("should have endpoints that are available on both", func() {
+
+		it.Before(func() {
+			service.Create(t, "ibuildthecloud/demo:v1")
+		})
+		it.After(func() {
+			service.Remove()
+			stagedService.Remove()
+		})
+
+		it("should have endpoints that are available on both with one app endpoint", func() {
 			stagedService = service.Stage("ibuildthecloud/demo:v3", "v3")
 
 			// Check the hostnames returned by Rio and Kubectl are equal
@@ -38,15 +39,42 @@ func endpointTests(t *testing.T, when spec.G, it spec.S) {
 
 			assert.Equal(t, "Hello World", service.GetEndpointResponse())
 			assert.Equal(t, "Hello World v3", stagedService.GetEndpointResponse())
-		})
-		it("should have the service app endpoint properly created", func() {
 
-			// Check the hostnames returned by Rio and Kubectl are equal
 			assert.Equal(t,
 				testutil.GetHostname(service.GetKubeAppEndpointURL()),
 				testutil.GetHostname(service.GetAppEndpointURL()),
 			)
+			assert.Equal(t,
+				testutil.GetHostname(service.GetAppEndpointURL()),
+				testutil.GetHostname(stagedService.GetAppEndpointURL()),
+			)
 			assert.Equal(t, "Hello World", service.GetAppEndpointResponse())
+		})
+	}, spec.Parallel())
+
+	when("a staged service is promoted", func() {
+		it.Before(func() {
+			service.Create(t, "ibuildthecloud/demo:v1")
+			stagedService = service.Stage("ibuildthecloud/demo:v3", "v3")
+			stagedService.Promote()
+		})
+		it.After(func() {
+			service.Remove()
+			stagedService.Remove()
+		})
+
+		it("should retain all revision endpoints with an app endpoint pointing to the new revision", func() {
+			assert.Equal(t, "Hello World", service.GetEndpointResponse())
+			assert.Equal(t, "Hello World v3", stagedService.GetEndpointResponse())
+			assert.Equal(t, 0, service.GetKubeCurrentWeight())
+			assert.Equal(t, 100, stagedService.GetKubeCurrentWeight())
+			assert.Equal(t, "Hello World v3", service.GetAppEndpointResponse())
+		})
+		it("should allow rolling back to the previous revision", func() {
+			service.Promote()
+			assert.Equal(t, 100, service.GetKubeCurrentWeight())
+			assert.Equal(t, 0, stagedService.GetKubeCurrentWeight())
+			assert.Equal(t, "Hello World", service.GetEndpointResponse())
 		})
 	}, spec.Parallel())
 }
--- a/tests/integration/run_test.go
+++ b/tests/integration/run_test.go
@@ -13,22 +13,63 @@ func runTests(t *testing.T, when spec.G, it spec.S) {
 
 	var service testutil.TestService
 
-	it.Before(func() {
-		service.Create(t, "nginx")
-	})
-
 	it.After(func() {
 		service.Remove()
 	})
 
 	when("rio run is called", func() {
-		it("should create a service with default specifications", func() {
-			runningPods := service.GetRunningPods()
+
+		it.Before(func() {
+			service.Create(t)
+		})
+
+		it("should create a service with default specifications that can scale up", func() {
 			assert.Equal(t, 1, service.GetAvailableReplicas(), "should have one available replica")
 			assert.Equal(t, 100, service.GetCurrentWeight())
 			assert.Equal(t, "nginx", service.GetImage())
-			assert.Contains(t, runningPods, service.App.Name)
-			assert.Contains(t, runningPods, "2/2")
+			assert.Contains(t, service.Logs(), "linkerd-init")
+			assert.Contains(t, service.Exec("env"), "KUBERNETES_SERVICE_PORT")
+			runningPods := service.GetRunningPods()
+			for _, pod := range runningPods {
+				assert.Contains(t, pod, service.App.Name)
+				assert.Contains(t, pod, "2/2")
+			}
+
+			service.GenerateLoad()
+			assert.Greater(t, service.GetAvailableReplicas(), 1, "should have more than 1 available replica")
+			runningPods = service.GetRunningPods()
+			assert.Greater(t, len(runningPods), 1)
 		})
-	})
+	}, spec.Parallel())
+
+	when("run a service with a fixed scale", func() {
+
+		it.Before(func() {
+			service.Create(t, "--scale", "3", "https://github.com/rancher/rio-demo")
+		})
+
+		it("should have the specified scale, no autoscaling, and be able to have its scale manually adjusted", func() {
+			assert.Equal(t, 3, service.GetAvailableReplicas(), "should have three available replicas")
+			assert.Equal(t, "Hi there, I'm running in Rio", service.GetAppEndpointResponse())
+
+			service.GenerateLoad()
+			assert.Equal(t, 3, service.GetAvailableReplicas(), "should have three available replicas")
+			runningPods := service.GetRunningPods()
+			assert.Len(t, runningPods, 3)
+			for _, pod := range runningPods {
+				assert.Contains(t, pod, service.App.Name)
+				assert.Contains(t, pod, "2/2")
+			}
+
+			service.Scale(1)
+			assert.Equal(t, 1, service.GetAvailableReplicas(), "should have one available replica")
+			assert.Equal(t, 1, service.GetScale())
+			runningPods = service.GetRunningPods()
+			assert.Len(t, runningPods, 1)
+			for _, pod := range runningPods {
+				assert.Contains(t, pod, service.App.Name)
+				assert.Contains(t, pod, "2/2")
+			}
+		})
+	}, spec.Parallel())
 }
--- a/tests/integration/scale_test.go
+++ b/tests/integration/scale_test.go
@@ -22,10 +22,11 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
 	})
 
 	when("scale is called on a service", func() {
-		it("should scale up", func() {
+		it("should scale up to 3", func() {
 			assert.Equal(t, 1, service.GetAvailableReplicas())
-			service.Scale(2)
-			assert.Equal(t, 2, service.GetAvailableReplicas())
+			service.Scale(3)
+			assert.Equal(t, 3, service.GetScale())
+			assert.Equal(t, 3, service.GetAvailableReplicas())
 			assert.Equal(t, service.GetKubeAvailableReplicas(), service.GetAvailableReplicas())
 			assert.True(t, service.PodsResponsesMatchAvailableReplicas("/name.html", service.GetAvailableReplicas()))
 		})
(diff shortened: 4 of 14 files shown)


izaac:
@rancher-max I think the private function `waitForReadyReplicas()` idea is a good one, could prevent some of the flakyness we previously faced with scaling.

tests/testutil/service.go
```diff
@@ -220,7 +239,22 @@ func (ts *TestService) GetRunningPods() string {
...
+		}
+	}
+	return runningPods
+}
+
+// GenerateLoad queries the endpoint multiple times in order to put load on the service
+func (ts *TestService) GenerateLoad() {
+	HeyCmd(GetHostname(ts.GetAppEndpointURL()), "90s", 130)
+	ts.reload()
+	ts.reloadApp()
```
cbron: Why not ts.GetEndpointUrl ? Why not pass in full url rather than just hostname ? 
Should probably make time and `c` params.

tests/testutil/testutil.go
```diff
@@ -24,11 +25,12 @@ func init() {
 }
 
 // Ensure CLI flag is passed, this way integration tests won't run during unit tests
-func PreCheck() {
-	runTests := flag.Bool("integration-tests", false, "must be provided to run the integration tests")
+func PreCheck(flagToCheck string) {
```
cbron: Passing in string like this is brittle, need to fix if we continue down the dual test path.

tests/validation/p1_run_test.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/rancher/rio/tests/testutil"
+)
+
+func p1RunTests(t *testing.T, when spec.G, it spec.S) {
```
cbron: I'm confused by this, I thought are integration tests were making up the p1 test cases ? This is mostly duplicated. Duplication is bad because it will increase our drone and general testing run time.

tests/integration/run_test.go
```diff
@@ -27,8 +27,10 @@ func runTests(t *testing.T, when spec.G, it spec.S) {
 			assert.Equal(t, 1, service.GetAvailableReplicas(), "should have one available replica")
 			assert.Equal(t, 100, service.GetCurrentWeight())
 			assert.Equal(t, "nginx", service.GetImage())
-			assert.Contains(t, runningPods, service.App.Name)
-			assert.Contains(t, runningPods, "2/2")
+			for _,pod := range runningPods {
```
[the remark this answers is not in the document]
rancher-max: I changed the `runningPods` function to return an array of the running pods, so now it will essentially returns something like:
```
[
"hi-6c9597ddb9-rl4nk   2/2     Running   0          28m",
"hi-6c9597ddb9-zjkwk   2/2     Running   0          34s",
"hi-6c9597ddb9-ct9mt   2/2     Running   0          34s"
]
```
...which is more useful to have because now we can validate that each pod retrieved has the service name and is running with 2/2. (Essentially, the scenario in this case is: "when we scale the service, all pods should be available")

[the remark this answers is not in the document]
cbron: Hmmm, I suppose so. Feels like you are testing Kubernetes itself now though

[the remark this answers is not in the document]
rancher-max: Maybe.. I think it's rio's flavor though, since it needs each pod to have both the linkerd proxy container and the app container to function properly

tests/testutil/service.go
rancher-max: `ts.GetEndpointUrl` returns the endpoint with the version included. I just prefer to use the same endpoint that would show up with a `rio ps` command in this function in case we ever use it to generate load on a service that has different versions with weights > 0.

Time and c will be params.

Using hostname because sometimes `hey` has issues with the https endpoint

tests/testutil/service.go
```diff
@@ -374,7 +408,7 @@ func (ts *TestService) waitForReadyService() error {
 	f := wait.ConditionFunc(func() (bool, error) {
 		err := ts.reload()
 		if err == nil {
-			if ts.Service.Status.DeploymentStatus != nil && ts.Service.Status.DeploymentStatus.AvailableReplicas > 0 {
+			if ts.Service.Status.DeploymentStatus != nil && ts.Service.Status.DeploymentStatus.AvailableReplicas > 0 && ts.Service.Status.DeploymentStatus.AvailableReplicas == ts.Service.Status.DeploymentStatus.Replicas {
```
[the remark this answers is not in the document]
rancher-max: I'm going to change this one. Stay tuned.

tests/testutil/testutil.go
rancher-max: Would it be better to have two functions here? `PreCheckIntegration` and `PreCheckValidation`?

cbron: Yea I guess, meh, feels gross either way. Or make them private methods in the suite files themselves.

tests/testutil/service.go
```diff
@@ -302,6 +397,24 @@ func (ts *TestService) GetKubeCurrentWeight() int {
 	return getRevisionWeight(app, ts.Version)
 }
 
+// WaitForScaleDown waits until either 5 minutes pass or a service has scaled down to its minimum value
```
izaac: Is 5 minutes because of the scale down to 0?

rancher-max: Yeah it usually takes approx. 3 minutes to go from 1 pod to 0, but I gave it some room in case this is used on a service that started with more pods.

tests/integration/domain_test.go
```diff
@@ -15,7 +15,7 @@ func domainTests(t *testing.T, when spec.G, it spec.S) {
 	var domain testutil.TestDomain
 
 	it.Before(func() {
-		service.Create(t, "nginx")
+		service.Create(t, "nginx", "-p", "80/http")
```
cbron: Wasn't it still going to default to this ?

tests/integration/endpoint_test.go
```diff
@@ -38,15 +39,42 @@ func endpointTests(t *testing.T, when spec.G, it spec.S) {
...
+
+		it("should retain all revision endpoints with an app endpoint pointing to the new revision", func() {
+			assert.Equal(t, "Hello World", service.GetEndpointResponse())
+			assert.Equal(t, "Hello World v3", stagedService.GetEndpointResponse())
+			assert.Equal(t, 0, service.GetKubeCurrentWeight())
+			assert.Equal(t, 100, stagedService.GetKubeCurrentWeight())
+			assert.Equal(t, "Hello World v3", service.GetAppEndpointResponse())
+		})
+		it("should allow rolling back to the previous revision", func() {
+			service.Promote("--rollout-increment", "25", "--rollout-interval", "2")
```
cbron: Doesn't this increment/interval screw up the test ? How does it get to 100 immediately.

tests/integration/run_test.go
```diff
@@ -13,22 +13,63 @@ func runTests(t *testing.T, when spec.G, it spec.S) {
...
+			assert.Contains(t, service.Logs(), "linkerd-init")
+			assert.Contains(t, service.Exec("env"), "KUBERNETES_SERVICE_PORT")
+			runningPods := service.GetRunningPods()
+			for _, pod := range runningPods {
+				assert.Contains(t, pod, service.App.Name)
+				assert.Contains(t, pod, "2/2")
+			}
+
+			service.GenerateLoad("80s", 120)
+			assert.Greater(t, service.GetAvailableReplicas(), 1, "should have more than 1 available replica")
```
cbron: Is there a risk of this not happening immediately ?

tests/testutil/service.go
```diff
@@ -11,21 +11,24 @@ import (
...
 	Service riov1.Service
+	Build   tektonv1alpha1.TaskRun
 	Version string
 	T       *testing.T
 }
 
 // Create generates a new rio service, named randomly in the testing namespace, and
 // returns a new TestService with it attached. Guarantees ready state but not live endpoint
-func (ts *TestService) Create(t *testing.T, source string) {
+func (ts *TestService) Create(t *testing.T, source string, args ...string) {
```
cbron: This is being done in https://github.com/rancher/rio/pull/548 also, make sure they match up.

tests/testutil/service.go
```diff
@@ -37,14 +40,27 @@ func (ts *TestService) Create(t *testing.T, source string) {
...
 	}
-	_, err := RioCmd([]string{"--namespace", testingNamespace, "run", "-p", "80/http", "-n", ts.AppName, source})
+	args = append(
+		[]string{"--namespace", testingNamespace, "run"},
+		append(args, "-n", ts.AppName, source)...)
+	_, err := RioCmd(args)
 	if err != nil {
 		ts.T.Fatalf("Failed to create service %s: %v", ts.Name, err.Error())
 	}
+	if source[0:4] == "http" && strings.Contains(source, "github") {
```
cbron: Is there a reason we need to check for both tekton and ready service ? Is it just to get more info. Can you double check also that you can't pass in a non scheme'd url string, like just `github.com/a/b`, as that would need to be taken into account here.

tests/testutil/service.go
```diff
@@ -114,6 +130,27 @@ func (ts *TestService) Weight(percentage int) {
...
 
+// RolloutWeight calls "rio weight --rollout=true --rollout-increment={increment} --rollout-interval={interval} ns/service:version={percentage}" on this service
+func (ts *TestService) RolloutWeight(increment int, interval int, percentage int) {
+	// Set increment and interval to their default values if a negative number is passed in
+	if increment < 0 {
+		increment = 5
+	}
+	if interval < 0 {
+		interval = 5
+	}
```
cbron: I don't see the point of this logic, if we want to normalize the values that should be done in the actual code. Also the tests are only ones passing this info in, no user input here.

tests/integration/run_test.go
```diff
@@ -13,22 +13,63 @@ func runTests(t *testing.T, when spec.G, it spec.S) {
...
-			assert.Contains(t, runningPods, "2/2")
+			assert.Contains(t, service.Logs(), "linkerd-init")
+			assert.Contains(t, service.Exec("env"), "KUBERNETES_SERVICE_PORT")
+			runningPods := service.GetRunningPods()
+			for _, pod := range runningPods {
+				assert.Contains(t, pod, service.App.Name)
+				assert.Contains(t, pod, "2/2")
+			}
+
+			service.GenerateLoad("80s", 120)
```
cbron: Is 80 seconds necessary ?

tests/testutil/service.go
```diff
@@ -220,7 +295,42 @@ func (ts *TestService) GetRunningPods() string {
...
+		if strings.Contains(pod, "Running") {
+			runningPods = append(runningPods, pod)
+		}
+	}
+	return runningPods
+}
+
+// GetResponseCounts takes two expected response strings and sends numRequests requests to the service's app endpoint.
+// If it gets a response other than one of the two specified, it throws a failure. Otherwise it returns individual counts of each response.
+func (ts *TestService) GetResponseCounts(response1 string, response2 string, numRequests int) (int, int) {
```
cbron: Interesting idea. Let's just make it a `[]string` instead.

tests/testutil/service.go
```diff
@@ -209,7 +284,7 @@ func (ts *TestService) GetImage() string {
 
 // GetRunningPods returns all running pods, separated by new lines, for this service's app
```
cbron: Believe this comment is no longer correct. The name isn't very descriptive either, I would assume it returns pod objects.

tests/testutil/service.go
cbron: Would it be easier to just send this info into `WaitForScale` ?

tests/testutil/service.go
```diff
@@ -388,12 +533,30 @@ func (ts *TestService) waitForReadyService() error {
...
+	return nil
+}
+
 // Wait until a service hits wanted number of available replicas, or error out
 func (ts *TestService) waitForAvailableReplicas(want int) error {
 	f := wait.ConditionFunc(func() (bool, error) {
 		err := ts.reload()
 		if err == nil {
-			if ts.GetAvailableReplicas() == want {
+			if ts.Service.Status.DeploymentStatus != nil && int(ts.Service.Status.DeploymentStatus.AvailableReplicas) == want {
```
cbron: Whats point of this change ? Isn't that exactly what `GetAvailableReplicas` does ?

tests/validation/p1_run_test.go
```diff
@@ -0,0 +1,46 @@
+package validation
```
cbron: Nitpick but this feels like scaling or autoscaling tests ? Not run tests.

tests/validation/suite_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"testing"
+
+	"github.com/sclevine/spec"
+	"github.com/sclevine/spec/report"
+
+	"github.com/rancher/rio/tests/testutil"
+)
+
+func TestMain(m *testing.M) {
+	testutil.PreCheck("validation-tests")
```
cbron: Const here

tests/testutil/testutil.go
cbron: Should make them const, or separate methods. Don't like the string still.

tests/testutil/service.go
rancher-max: That PR did this in a much more elegant way. I'll update this to do the same. Need to figure out a better way to default port to 80/http. It seems to work on our docker images but fails on GH builds. Potential solution is including the argument if the source isn't from GH. I'll do that and we can change if we see issues.

tests/integration/run_test.go
rancher-max: Yes... It happens quickly usually when tests aren't run in parallel, but having more tests == more running in parallel == more pressure on CPU == greater timetoscale :(

tests/integration/endpoint_test.go
rancher-max: The promote function waits for weight to be 100%. I'm glad you brought this up. I think maybe for integration tests it makes sense to not have a rollout, and we can (and do) validate rollout in other tests. I'll change this one.

tests/testutil/service.go
rancher-max: non scheme'd url string assumes a docker image (tested and confirmed: `v0: test(ImagePullBackOff/Back-off pulling image "github.com/rancher/rio-demo")`

I'm checking tekton so as not to have to generically increase ready service timeout. When a build happens (passing in GH repo), the time until the service is ready is much much larger, especially with the tests running in parallel, so I was noticing flakiness.

tests/testutil/service.go
rancher-max: I changed this like 8 times trying to solve flakiness. This was a mistake :(

tests/testutil/service.go
rancher-max: I tried that originally. `WaitForScale` doesn't quite do what we want though because it returns the `Spec.Scale` value, which is still 1 even if the service scales down to 0.

tests/integration/run_test.go
rancher-max: Not for the autoscaling in a single test, but for the CPU of the cluster to actually apply it. I'm not sure how to make this one less flaky or take less time. I'm open to all ideas because I don't like it either.

cbron: Follow up question, if it only takes 5 seconds does the hey command get shut down or does it try to keep running against a removed service ?

cbron: Maybe you could do a 10 second load and then check, and loop that until it works.

rancher-max:
@cbron See latest commits. I think they address most of your comments. The loop idea is interesting and something I can look into, but I don't think it's necessary. I just hardcoded it to 30s with 10 * maxscale for concurrent requests. I think this should be enough load in all test cases and won't take too long. Happy to change it to loop method with a timeout though if you think that's better.

However, checkout the failure in CI. It's failing on services not being ready:
`--- FAIL: TestSuite/integration_suite/endpoint/a_staged_service_is_promoted/should_retain_all_revision_endpoints_with_an_app_endpoint_pointing_to_the_new_revision (120.43s)`
`service.go:45: service testing-ns/mhcsb:v0 never reached ready status`
`--- FAIL: TestSuite/integration_suite/endpoint/a_staged_service_is_promoted/should_allow_rolling_back_to_the_previous_revision (120.28s)`
`service.go:45: service testing-ns/rckih:v0 never reached ready status`

izaac:
@rancher-max I wonder If we're reaching a limit in the CI cluster we're using for testing maybe there are too many replicas created at the same time? I think it would all depend on the host, but that's unknown to me in Drone.

rancher-max:
The issue was my fault. I left some code after fixing my changes to service to not require port to be passed in. After cleaning that code up, it looks like those flaky tests were resolved.

tests/integration/scale_test.go
```diff
@@ -34,23 +35,11 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
...
-			assert.Equal(t, 3, service.GetScale())
-		})
-		it("should scale to 5", func() {
-			assert.Equal(t, 1, service.GetAvailableReplicas())
-			service.Scale(5)
-			assert.Equal(t, 5, service.GetAvailableReplicas())
-			assert.Equal(t, 5, service.GetScale())
-		})
-		it("should scale to 10", func() {
+		it("should scale above 10", func() {
```
izaac: The spec description should match the actual scale value tested.

tests/integration/run_test.go
cbron: OK if it gets flakey just add a wait method.

tests/testutil/service.go
```diff
@@ -111,6 +138,20 @@ func (ts *TestService) Weight(percentage int) {
 	}
 }
 
+// RolloutWeight calls "rio weight --rollout=true --rollout-increment={increment} --rollout-interval={interval} ns/service:version={percentage}" on this service
+func (ts *TestService) RolloutWeight(increment int, interval int, percentage int) {
```
cbron: Can we just use func `weight` and pass in params ? It'd be good to keep funcs and cli command as close to 1:1 as possible.

tests/validation/suite_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+func TestMain(m *testing.M) {
+	testutil.ValidationPreCheck()
+	os.Exit(m.Run())
+}
+
+func TestSuite(t *testing.T) {
+	suite := spec.New("validation suite", spec.Report(report.Terminal{}))
+	specs := map[string]func(t *testing.T, when spec.G, it spec.S){
+		"p1run":    p1AutoscaleTests,
```
cbron: I wonder if we shouldn't use `p1`, as that is an internal term. Community members have no idea what that means.

tests/testutil/service.go
rancher-max: Resolved! May want to take a look at the changes I made to fix that. We already had a `func Weight` so I just updated that and its uses to specify rollout or not.

PR #581, opened 2019-10-19
Removing unused python dependencies from CI Dockerfile

izaac:
We no longer use python for test execution on CI

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -5,8 +5,7 @@ ENV ARCH $DAPPER_HOST_ARCH
 ENV K3S_VERSION v0.9.1
 ENV KUBECTL_VERSION v1.15.0
 
-RUN apk -U add bash git gcc musl-dev docker vim less file curl wget ca-certificates python3 py3-pip python3-dev openssl-dev libffi-dev jq
-RUN pip3 install 'tox==3.6.0'
+RUN apk -U add bash git gcc musl-dev docker vim less file curl wget ca-certificates
 RUN go get -d golang.org/x/lint/golint && \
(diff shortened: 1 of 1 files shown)


Dockerfile.dapper
```diff
@@ -5,8 +5,7 @@ ENV ARCH $DAPPER_HOST_ARCH
 ENV K3S_VERSION v0.9.1
 ENV KUBECTL_VERSION v1.15.0
 
-RUN apk -U add bash git gcc musl-dev docker vim less file curl wget ca-certificates python3 py3-pip python3-dev openssl-dev libffi-dev jq
-RUN pip3 install 'tox==3.6.0'
+RUN apk -U add bash git gcc musl-dev docker vim less file curl wget ca-certificates
```
cbron: Are we confident that `openssl-dev libffi-dev jq` aren't needed for rio ? This file is used for building too, not just running tests.

izaac: @cbron  We must add a step for the build process as part of the CI's if not already there. 

With `jq` I'm pretty confident is fine to remove, but besides running `make` locally and what CI's reported that `rio-controller`'s was successfully built I'm not sure what else to check for the `*-dev` ones.

izaac: FWIW browsing the Dockerfile log I just see those python dependencies added when the python test framework was introduced:

https://github.com/rancher/rio/commit/653c29eece314ad3f0b175a4c18bb90ee96322a7#diff-73040b7579d3747d271b2329159c2f83

cbron: openssl is the one I am curious about, I think we can probably try this and see what happens.

PR #585, opened 2019-10-21
Adding kube validations for externalservices

izaac:
This is part of #559 Add kubectl validations to the test framework

### DIFF
--- a/tests/testutil/externalservice.go
+++ b/tests/testutil/externalservice.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	riov1 "github.com/rancher/rio/pkg/apis/rio.cattle.io/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type TestExternalService struct {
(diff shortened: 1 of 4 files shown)


tests/testutil/externalservice.go
```diff
@@ -72,6 +73,42 @@ func (es *TestExternalService) GetFQDN() string {
...
+	}
+	if len(endpointsList.Items) == 0 {
+		return ""
+	}
+	if len(endpointsList.Items[0].Subsets) == 0 {
+		return ""
+	}
+	if len(endpointsList.Items[0].Subsets[0].Addresses) == 0 {
+		return ""
+	}
```
rancher-max: You can combine these 3 conditions into 1 here

tests/testutil/externalservice.go
```diff
@@ -72,6 +73,42 @@ func (es *TestExternalService) GetFQDN() string {
...
+// GetKubeFQDN retrieves the external service FQDN value using the kubernetes clientset
+func (es *TestExternalService) GetKubeFQDN() string {
+	clientset := GetKubeClient()
+	serviceList, err := clientset.CoreV1().
+		Services(testingNamespace).
+		List(metav1.ListOptions{LabelSelector: "rio.cattle.io/service=" + es.ExternalService.Name})
+	if err != nil {
+		es.T.Fatalf(err.Error())
+	}
+	if !(len(serviceList.Items) > 0) {
```
rancher-max: This is fine, but just for consistency I think it should check for `== 0` as well

izaac:
linter is failing with the latest refactor

scripts/test
```diff
@@ -1,14 +1,27 @@
...
-cd $(dirname $0)/..
+cd "$(dirname $0)"/..
 
 fatal()
 {
     echo "[ERROR] " "$@"
     exit 1
 }
 
+waitUntilk3sIsReady()
```
cbron: This may not be necessary anymore, there is now an install checker.

scripts/test
```diff
@@ -20,10 +33,12 @@ if [[ ${ARCH} == amd64 ]]; then
     go test -cover -tags=test ./...
 
     k3s server >/dev/null 2>&1 &
-    sleep 10
-
     export KUBECONFIG=/etc/rancher/k3s/k3s.yaml
-    export PATH=$(pwd)/bin:$PATH
+    waitUntilk3sIsReady
+
+    k3s kubectl apply -f ./scripts/rbac.yaml
```
cbron: This can be removed

scripts/test
```diff
@@ -20,10 +33,12 @@ if [[ ${ARCH} == amd64 ]]; then
...
+    PATH=$(pwd)/bin:$PATH
```
cbron: Any reason for the split ?

tests/testutil/externalservice.go
```diff
@@ -72,6 +73,38 @@ func (es *TestExternalService) GetFQDN() string {
...
+// GetKubeFQDN retrieves the external service FQDN value using the kubernetes clientset
+func (es *TestExternalService) GetKubeFQDN() string {
+	clientset := GetKubeClient()
+	serviceList, err := clientset.CoreV1().
+		Services(testingNamespace).
+		List(metav1.ListOptions{LabelSelector: "rio.cattle.io/service=" + es.ExternalService.Name})
+	if err != nil {
+		es.T.Fatalf(err.Error())
+	}
+	if len(serviceList.Items) == 0 {
```
cbron: Might make the above into a helper and then have both these methods just call that so it's not duplicated.

scripts/test
izaac: The bash linter recommended separating the assignment from the export to prevent value masking.

tests/testutil/externalservice.go
izaac: Those two are different types, one Endpoints and the other Services what would be the preferred way to do that, type checking?

cbron: ahh, true

izaac:
> Will need to be rebased now.

Rebased

PR #586, opened 2019-10-21
Falling back to default ~/.kube/config if no KUBECONFIG env is set for client-go

izaac:
This is the default rio/kubectl behavior, we should do the same for the client-go

### DIFF
--- a/tests/testutil/testutil.go
+++ b/tests/testutil/testutil.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tests/testutil/testutil.go
```diff
@@ -132,11 +133,24 @@ func stringInSlice(a string, list []string) bool {
 // KUBECONFIG env value
 func GetKubeClient() *kubernetes.Clientset {
 	kubeConfigENV := os.Getenv("KUBECONFIG")
+	if kubeConfigENV == "" {
+		if home := homeDir(); home != "" {
+			kubeConfigENV= filepath.Join(home, ".kube", "config")
+		}
```
rancher-max: else maybe throw a descriptive error here to let the tester know to set the KUBECONFIG environment variable

izaac: I also thought on exiting with a terminal error code (Exit 1) if this happens.

PR #589, opened 2019-10-23
Allows logs to lookup builds automatically

cbron:


cli/cmd/logs/logs.go
```diff
@@ -40,33 +40,36 @@ func (l *Logs) Run(ctx *clicontext.CLIContext) error {
...
+	if obj.Object == nil {
+		return nil, errors.New("No object found")
 	}
 
-	tail := int64(l.N_Tail)
-	config.TailLines = &tail
+	// handle both build and service lookups
+	var lookupID string
+	if obj.Type == clitypes.BuildType {
+		lookupID = strings.TrimLeft(obj.Name, "taskrun/default/")
```
StrongMonkey: nit: lookupId can be found in object.Name

StrongMonkey: I think this should be changed to use object.Name, default shouldn't be hard-coded.

PR #606, opened 2019-10-25
allow version to be specd in service name on run

cbron:
Allows a user to run: `rio run -n svc:v1 nginx` and have version correctly set


cli/cmd/create/create.go
```diff
@@ -97,6 +98,9 @@ func (c *Create) RunCallback(ctx *clicontext.CLIContext, cb func(service *riov1.
 	}
 
 	service.Namespace, service.Name = stack.NamespaceAndName(ctx, service.Name)
+	if strings.Contains(service.Name, ":") {
```
StrongMonkey: nit: `service.Name, service.Spec.Version = kv.Split(service.Name, ":")` is enough

PR #629, opened 2019-10-28
Fix rbac tests, make install experience better

StrongMonkey:


tests/integration/scale_test.go
```diff
@@ -37,11 +37,11 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
 			assert.Equal(t, 0, service.GetAvailableReplicas())
 			assert.Equal(t, 0, service.GetScale())
 		})
-		it("should scale to 11", func() {
+		it("should scale to 2", func() {
 			assert.Equal(t, 1, service.GetAvailableReplicas())
-			service.Scale(11)
-			assert.Equal(t, 11, service.GetAvailableReplicas())
-			assert.Equal(t, 11, service.GetScale())
```
cbron: Is this temporary ? Or for some reason ?

StrongMonkey: @cbron  I don't want to launch 11 pods, I think 2 pods should be enough to test the scale functionality though...

StrongMonkey: Launch 11 pods make tests pretty slow...

cbron: If thats the case then @rancher-max can you move the 11 scale test to validation.

Issue #616, opened 2019-10-28
rio ps -q option returns error


izaac:
**Describe the bug**
`rio ps` has a `-q` option to only display names.
```bash
rio ps --help
NAME:
   rio ps - List services

USAGE:
   rio ps [OPTIONS]

DESCRIPTION:
   To view all rio services, run `rio ps`

OPTIONS:
   --quiet, -q     Only display Names
   --format value  'json' or 'yaml' or Custom format: '{{.Name}} {{.Obj.Name}}' [$FORMAT]
   --all, -a       print all resources, including router and externalservice
```

But it returns: FATA[0000] template: :1:2: executing "" at <.Name>: can't evaluate field Name in type ps.ServiceData

**To Reproduce**
1. install rio, spin up a service.
2. do a `rio ps -a` or `rio ps -q -a` and just a `rio ps -a`
3. `rio ps -a` works only, using `-q` returns FATA error.

**Expected behavior**

Returned output as described in `rio ps --help`

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2-k3s.1", GitCommit:"b8b17ba55f20e590df507fce333dfee13ab438c6", GitTreeState:"clean", BuildDate:"2019-10-16T05:17Z", GoVersion:"go1.13.3", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
rio version v0.6.0-alpha.1 (7678d67d)
```

StrongMonkey:
Available to test in v0.6.0-alpha2.

rancher-max:
Fixed in v0.6.0-alpha.2

Issue #630, opened 2019-10-29
Default ClusterRoles for RBAC give error on rio ps -a


rancher-max:
**Describe the bug**
When running `rio ps -a` while impersonating rio-standard, rio-readonly, or rio-privileged there is a panic error.

**To Reproduce**
1. `kubectl apply -f crb.yaml`
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: rio-standard-crb
subjects:
- kind: Group
  name: rio:standard
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole #this must be Role or ClusterRole
  name: rio-standard # this must match the name of the Role or ClusterRole you wish to bind to
  apiGroup: rbac.authorization.k8s.io
```
2. Update kubeconfig to use the rio-standard clusterrole:
```
...
users:
- name: default
  user:
    password: fakepass
    username: fakeuser
    as: rio-standard
    as-groups:
    - rio:standard
```
3. Run `rio ps -a`

**Expected behavior**
Should show available routers and externalservices as well as services.

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-19T13:57:45Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2-k3s.1", GitCommit:"b8b17ba55f20e590df507fce333dfee13ab438c6", GitTreeState:"clean", BuildDate:"2019-10-16T05:17Z", GoVersion:"go1.13.3", Compiler:"gc", Platform:"linux/amd64"}
```

**Rio version**: `v0.6.0-alpha.1`

StrongMonkey:
Available to test in v0.6.0-alpha2.

rancher-max:
Working in v0.6.0-alpha.2

Issue #608, opened 2019-10-26
rio attach now showing pod stdout


izaac:
**Describe the bug**
doing a `rio attach <service>` is now showing the container stdout as in 0.5.0

**To Reproduce**
1. Deploy a new service `rio run --name attachtest izaac/attachtest:v1`
2. `rio attach attachtest`

Nothing is shown after the service is ready, this container shows the `date` every 5 seconds to stdout.

**Expected behavior**

Attach Should show output from the pod in this case output of the running bash script running `date` every 5 seconds.

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
v1.14.7-gke.10
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.6.0-alpha.1 (7678d67d)
Rio CLI Version: v0.6.0-alpha.1 (7678d67d)
```

StrongMonkey:
Available to test in v0.6.0-alpha2.

StrongMonkey:
@rancher-max Also make sure `rio run --it ${image}` also works. This by default will launch the container and attach to its main process.

rancher-max:
Fixed in v0.6.0-alpha.2.
Also verified `rio run -n attach2 -it izaac/attachtest:v1` and working as expected.

izaac:
This is working for me too in rio `0.6.0-alpha.2`, I've tested in k3s 0.10.1 and EKS clusters.

Issue #603, opened 2019-10-25
Routes are returning HTTP 503


izaac:
**Describe the bug**
Create a service and add a router in the mesh. The endpoint works fine but the router is returning HTTP 503.

**To Reproduce**
1. Create a service `rio run -n testcontainer -p 80 sangeetha/mytestcontainer` 
2. Wait until the endpoint is available.
3. Create a route: `rio router add route1/to-tsrv-v0 to default/testcontainer`
4. Wait until the router is generated. `rio router ls`

**Expected behavior**

Router Endpoint should route the requests to the app deployed.

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
Tried both:
```
v1.14.7-gke.10
v1.16.2-k3s.1
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.6.0-alpha.1 (7678d67d)
Rio CLI Version: v0.6.0-alpha.1 (7678d67d)
```
**Additional context**
`rio systemlog`
output:
```
Nothing relevant.
```

StrongMonkey:
Available to test in v0.6.0-alpha2.

rancher-max:
Working in v0.6.0-alpha.2.
Syntax might've changed a little bit: `rio router add route1/to-srv-v1 to testcontainer:v1`
Where `v1` is the version shown with `rio ps`

To do this in a different namespace: 
`rio -n issue-ns run -n testcontainer -p 80 sangeetha/mytestcontainer`
`rio ps` --> `...   testcontainer/53265b61     ... `
`rio -n issue-ns router add route1/to-srv-v1 to testcontainer:53265b61`

Note also with this container, due to nginx, the route will give an nginx 404 because it doesn't know which backend /to-srv-v1 should go to. Feel free to experiment with other containers :)

PR #636, opened 2019-10-29
Fix multiple cli and backend bugs

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rio/issues/616 https://github.com/rancher/rio/issues/608 https://github.com/rancher/rio/issues/603 https://github.com/rancher/rio/issues/602 https://github.com/rancher/rio/issues/630

cli/pkg/tables/service.go
```diff
@@ -16,20 +16,19 @@ import (
...
 		{"WEIGHT", "{{.Service | formatWeight}}"},
 		{"CREATED", "{{.Service.CreationTimestamp | ago}}"},
 		{"DETAIL", "{{serviceDetail .Service .Pod}}"},
 	}, cfg)
 
 	writer.AddFormatFunc("image", FormatImage)
 	writer.AddFormatFunc("scale", formatRevisionScale)
-	writer.AddFormatFunc("app", app)
+	writer.AddFormatFunc("appAndVersion", appAndVersion)
 	writer.AddFormatFunc("version", version)
```
cbron: Do we still need version ?

modules/rdns/controllers/service/handler.go
```diff
@@ -183,7 +183,7 @@ func (h *handler) staticAddress() ([]adminv1.Address, error) {
 func (h *handler) getAddresses(svc *corev1.Service) ([]adminv1.Address, bool, error) {
 	result, err := h.staticAddress()
 	if len(result) > 0 || err != nil {
-		return result, false, err
+		return result, true, err
```
cbron: this is ok ?

StrongMonkey: This is the fix for RKE bug, when IP-address is set nodeport must be returned to true.

cli/pkg/tables/service.go
StrongMonkey: whoops, don't need it anymore

Issue #617, opened 2019-10-28
rio continuous delivery does not update service image


rancher-max:
**Describe the bug**
CD through Github builds everything properly on commits to the specified branch, but does not update the `.spec.image` value on the rio service to use the newly built image

**To Reproduce**
1. Run a service from Github
2. Push a commit to that Github repo

**Expected behavior**
It should build the image with that new commit and update the service so that it becomes available from that app version.

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-19T13:57:45Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.7-gke.10", GitCommit:"8cea5f8ae165065f0d35e5de5dfa2f73617f02d1", GitTreeState:"clean", BuildDate:"2019-10-05T00:08:10Z", GoVersion:"go1.12.9b4", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.6.0-alpha.1 (7678d67d)
Rio CLI Version: v0.6.0-alpha.1 (7678d67d)
```
**Additional context**
Snippets from `rio inspect <service>`:
```
...
spec:
  app: cd
  build:
    branch: k3s-branch
    repo: https://github.com/rancher-max/rio-demo
    revision: 6a724865960b7cb6aebc54500e35b6cfc0159628
  image: localhost:5442/default-lucid-buck4:c87b3
  imagePullPolicy: IfNotPresent
...
```
Notice how the `.spec.build.revision` value and the `.spec.image` suffix don't line up. If I manually change the image to end with `6a724` then it has the correct change.

StrongMonkey:
Available to test in v0.6.0-alpha2.

rancher-max:
Working in v0.6.0-alpha.2

PR #639, opened 2019-10-30
Fix new templating

StrongMonkey:


StrongMonkey:
This PR has new logic for generating services from build template if new commit is observed in git. We use a template generating controller to generate service right now.

StrongMonkey:
fix https://github.com/rancher/rio/issues/617

PR #643, opened 2019-10-30
Riofiles will modify resources

daxmc99:
Riofiles will modify resources

This changes modifies the apply logic to allow for a set of previously
seen GVKs to be passed into the Apply logic.

This allows for the stack to delete objects that are no longer present
in the Riofile on the next `rio up` run.

### DIFF
--- a/cli/cmd/up/up.go
+++ b/cli/cmd/up/up.go
@@ -14,8 +14,11 @@ import (
 	riov1 "github.com/rancher/rio/pkg/apis/rio.cattle.io/v1"
 	"github.com/rancher/rio/pkg/constants"
 	"github.com/rancher/rio/pkg/stack"
+	"github.com/rancher/wrangler/pkg/gvk"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 type Up struct {
@@ -30,6 +33,7 @@ type Up struct {
 const (
 	defaultRiofile       = "Riofile"
 	defaultRiofileAnswer = "Riofile-answers"
+	stackLabel           = "rio.cattle.io/stack"
 )
 
 func (u *Up) Run(c *clicontext.CLIContext) error {
@@ -46,7 +50,7 @@ func (u *Up) Run(c *clicontext.CLIContext) error {
 		return err
 	}
 
-	if err := u.updateStack(content, answer, c); err != nil {
+	if err = u.saveStack(content, answer, c); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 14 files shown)


daxmc99:
Wrangler PR https://github.com/rancher/wrangler/pull/45/files

tests/testutil/riofile.go
```diff
@@ -37,6 +37,24 @@ func (trf *TestRiofile) Up(t *testing.T, filename string) {
 	}
 }
 
+// Bring up a riofile by fixture file
+func (trf *TestRiofile) UpNonUnique(t *testing.T, filename, stackName string) {
+	trf.T = t
+	trf.Name = fmt.Sprintf("stack/%s/%s", testingNamespace, stackName)
```
cbron: Should put the rest of method below this into a common method that both of them call. Or just add `stackName` as var and if empty string we do the random generation.

daxmc99: Changed to perform this behavior

cli/cmd/up/up.go
```diff
@@ -142,10 +146,35 @@ func (u *Up) up(content string, answers map[string]string, s *riov1.Stack, c *cl
...
+		knowngvks = s.Spec.AdditionalGroupVersionKinds
+	}
+
+	err = c.Apply.WithListerNamespace(c.GetSetNamespace()).WithDefaultNamespace(c.GetSetNamespace()).WithOwner(s).WithSetOwnerReference(true, true).WithGVK(knowngvks).WithDynamicLookup().ApplyObjects(objs...)
+	if err != nil {
+		return err
+	}
+
+	stackToUpdate := s.DeepCopy()
+	stackToUpdate.Spec.AdditionalGroupVersionKinds = gvks
```
StrongMonkey: Extra assignment

cli/cmd/up/up.go
```diff
@@ -185,3 +214,20 @@ func getCurrentDir() string {
 	dir := filepath.Base(workingDir)
 	return strings.ToLower(dir)
 }
+
+func setObjLabels(objs []runtime.Object, s *riov1.Stack) ([]runtime.Object, error) {
+
+	for _, obj := range objs {
+
+		obj := obj.(metav1.Object)
```
StrongMonkey: To do a safe programming make sure to do
`obj, err := obj.(metav1.Object)` in case obj is not a metav1.Object and code won't panic.

cli/cmd/up/up.go
```diff
@@ -185,3 +214,20 @@ func getCurrentDir() string {
...
+		labels := obj.GetLabels()
+		if labels == nil {
+			labels = make(map[string]string)
+		}
+		// match debugID of stack
+		labels[stackLabel] = s.Namespace + "/" + s.Name
```
StrongMonkey: Don't have to add s.Namespace. Stack.Namespace can be determined by child.Namespace.
It could be 
`labels[stackLabel] =  s.Name`

StrongMonkey: Also I think we need to do
` obj.SetLabels(labels)`
Maybe I am wrong, not sure if labels are actually saved.

tests/integration/riofile_test.go
```diff
@@ -53,5 +55,43 @@ func riofileTests(t *testing.T, when spec.G, it spec.S) {
...
+			assert.Equal(t, strings.Trim(riofile.Readfile(), "\n"), strings.Trim(riofile.ExportStack(), "\n"), "should have stack export be same as original file")
+
+			// check frontend service came up
+			guestBook := testutil.CreateRoute(t, "guestbook", "frontend", "to")
+			defer guestBook.Remove()
+			guestBook.GetEndpointResponse()
+			//nginix service
+			nginxsvc := testutil.GetService(t, "nginx", "v0")
+			nginxsvc.GetEndpointResponse()
+
```
rancher-max: Can we add assertions to both of these `GetEndpointResponse()` calls just in case of false positives?

tests/integration/riofile_test.go
```diff
@@ -53,5 +55,43 @@ func riofileTests(t *testing.T, when spec.G, it spec.S) {
...
+			//assert k8s services came up outside of riofile
+			riofile.Up(t, "riofile-with-K8s-manifests-after.yaml", riofileName)
+			defer riofile.Remove()
+
+			//check that services/pods/deployments are removed
+			time.Sleep(30 * time.Second)
```
rancher-max: Is there a way that we can wait for the services/pods/deployments to be removed with a timeout instead of sleeping? Would hopefully fix potential flakiness on less powerful machines with tests running in parallel.

scripts/build
```diff
@@ -18,4 +18,4 @@ if [ "$CROSS" = "true" ] && [ "$ARCH" = "amd64" ]; then
     GOOS=darwin go build -ldflags "$LINKFLAGS" -o bin/rio-darwin ./cli
     GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/rio-windows ./cli
 fi
-CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/rio-controller
+GOOS=linux CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/rio-controller
```
cbron: This will prob break our local scripts to build controller, but we can work around that for now

tests/integration/riofile_test.go
cbron: +1, something like `GetService` for removals

tests/testutil/router.go
```diff
@@ -19,6 +19,26 @@ type TestRoute struct {
 	T      *testing.T
 }
 
+// create an arbitrary router to a service
+// ie "rio router add
+func CreateRoute(t *testing.T, name, svc, action string) TestRoute {
```
cbron: Whats difference between `CreateRoute` and `Add` ?

scripts/build
daxmc99: Yeah, if we have people using script build to run the controller locally but I think most of the scripted flows want to build the controller binary in order to load it into a docker image.

tests/testutil/router.go
daxmc99: Add requires a router to exist. This one is creating the router as well.

tests/integration/riofile_test.go
daxmc99: Updated to address this was a new function `WaitForNoResponse`

tests/testutil/router.go
cbron: Main reason for that is so we can have Spec handle the cleanup instead of using `defer` inline. If we are going to go this route, Create should at least just create a `TestRoute` and then call Add on it. The different signatures are also somewhat confusing, we should keep the order/names same as actual CLI call.

Issue #645, opened 2019-10-30
Delete error message in CLI when config is not found


StrongMonkey:
The error message to detect whether rio is installed into the cluster is kind of buggy. We should remove it for now.

davidnuzik:
Removed "big block" of content shown if config not found now. Should be fixed now.

rancher-max:
The big block of content is gone in v0.6.0. It has been replaced by a FATAL error as described in the referenced issue above. We can enhance that in the future.

PR #647, opened 2019-10-30
Don't prompt error message about install

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/645

PR #653, opened 2019-10-31
Add attach and log tests and update service readystate check

### LINKED ISSUE #693: Automation: Giving tailing rio commands a timeout
This is useful for commands that tail by default, i.e `rio attach` and `rio logs`. Then tests for these can be added.

rancher-max:
Closes #693

### DIFF
--- a/tests/testutil/service.go
+++ b/tests/testutil/service.go
@@ -122,6 +122,15 @@ func (ts *TestService) Remove() {
 	}
 }
 
+// Attach attaches to the service: `rio --namespace testing-ns attach <service name>` and appends each line of output to an array
+func (ts *TestService) Attach() []string {
+	results, err := RioCmdWithTail(10, []string{"attach", ts.Name})
+	if err != nil {
+		ts.T.Fatalf("Failed to get attach output: %v", err.Error())
+	}
+	return results
+}
+
 // Call "rio scale ns/service={scaleTo}"
 func (ts *TestService) Scale(scaleTo int) {
 	_, err := RioCmd([]string{"scale", fmt.Sprintf("%s=%d", ts.Name, scaleTo)})
@@ -140,8 +149,8 @@ func (ts *TestService) Scale(scaleTo int) {
 	}
 }
 
-// Weight calls "rio weight {args} service_name={weightSpec}" on this service.
-func (ts *TestService) Weight(weightSpec int, args ...string) {
+// WeightWithoutWaiting calls "rio weight {args} service_name={weightSpec}" on this service.
+func (ts *TestService) WeightWithoutWaiting(weightSpec int, args ...string) {
 	args = append(
 		[]string{"weight"},
 		append(args, fmt.Sprintf("%s=%d", ts.Name, weightSpec))...)
@@ -149,14 +158,19 @@ func (ts *TestService) Weight(weightSpec int, args ...string) {
 	if err != nil {
 		ts.T.Fatalf("weight command failed:  %v", err.Error())
 	}
+}
+
+// Weight calls "rio weight {args} service_name={weightSpec}" on this service and waits until the service weight reaches the desired value.
+func (ts *TestService) Weight(weightSpec int, args ...string) {
+	ts.WeightWithoutWaiting(weightSpec, args...)
 	paused := false
 	for _, a := range args {
-		if strings.Contains(a, "pause") {
+		if a == "pause" || a == "pause=true" {
 			paused = true
 		}
 	}
 	if !paused {
-		err = ts.waitForWeight(weightSpec)
+		err := ts.waitForWeight(weightSpec)
 		if err != nil {
 			ts.T.Fatal(err.Error())
 		}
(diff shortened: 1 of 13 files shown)


Dockerfile.dapper
```diff
@@ -2,8 +2,8 @@ FROM golang:1.13.3-alpine3.10
 
 ARG DAPPER_HOST_ARCH
 ENV ARCH $DAPPER_HOST_ARCH
-ENV K3S_VERSION v0.9.1
-ENV KUBECTL_VERSION v1.15.5
+ENV K3S_VERSION v0.10.0
```
cbron: We don't want to bump this yet

tests/testutil/service.go
```diff
@@ -198,13 +207,15 @@ func (ts *TestService) Promote(args ...string) {
...
 
 // Logs calls "rio logs ns/service" on this service
-func (ts *TestService) Logs(args ...string) string {
-	args = append([]string{"logs"}, append(args, ts.Name, "-a")...)
-	out, err := RioCmd(args)
+func (ts *TestService) Logs(args ...string) []string {
+	args = append(
+		[]string{"logs"},
+		append(args, ts.Name)...)
+	results, err := RioCmdWithTail(10, args)
```
cbron: So this follow logs for 10 seconds and then kills it ? Ok for now but feels like a stopgap solution. Hopefully in next month or so we will make a log dump option instead of always following.

tests/validation/suite_test.go
```diff
@@ -12,6 +12,7 @@ import (
 
 func TestMain(m *testing.M) {
 	testutil.ValidationPreCheck()
+	testutil.KubectlCmd([]string{"create", "namespace", testutil.TestingNamespace})
```
cbron: Whats this for ? If it's for previous flakey tests that should be resolved already.

tests/validation/weight_test.go
```diff
@@ -26,17 +26,16 @@ func weightTests(t *testing.T, when spec.G, it spec.S) {
 	})
 
 	when("a staged service incrementally rolls out weight", func() {
-		it("should slowly increase weight on the staged service and decrease current service weight", func() {
-			// This test is known to fail until https://github.com/rancher/rio/issues/577 is resolved.
+		it("should slowly increase weight on the staged service and leave service weight unchanged", func() {
 			// The time from rollout to obtaining the current weight, without Sleep, is 2 seconds.
 			// Sleeping 8 seconds here with a rollout-interval of 4 seconds to guarantee 2 rollout ticks, plus the initial tick, with 2 seconds to spare.
-			stagedService.Weight(80, "--increment=10", "--interval=4s")
+			stagedService.Weight(80, "--increment=10", "--interval=4", "--pause=false")
```
cbron: Shouldn't be necessary but we can leave.

cbron:
Can you double check you have rebased upstream/master too.

tests/validation/suite_test.go
rancher-max: Yes this is to resolve some of the flakiness. Does #701 resolve? I noticed that tests sometimes fail in a new cluster (or cluster that just doesn't have the `testutil.TestingNamespace` namespace) due to some sort of race in the parallel tests all trying to create it at once.

tests/validation/weight_test.go
rancher-max: Yeah I put the pause because of this snippet specifying to wait for the weight or not in `service.Weight`: 
```
	paused := false
	for _, a := range args {
		if strings.Contains(a, "pause") {
			paused = true
		}
	}
	if !paused {
		err = ts.waitForWeight(weightSpec)
		if err != nil {
			ts.T.Fatal(err.Error())
		}
	}
```

I felt writing another function (`RolloutWeight` or something) would be redundant, and I didn't like adding args and stripping them out. I agree it's meh though and maybe later we can come up with a better solution.

tests/validation/suite_test.go
cbron: It was an earlier PR that fixed it, should be good to remove this. If it acts up in future we can re-add.

tests/validation/weight_test.go
cbron: bleh, that snipped is flawed (guess who wrote it), can you add a check there where it tracks int through loop and checks if curr +1 == "true"

rancher-max: I liked the flaw! It was exploitable in a good way :)
For weight, we have to either just the flag: `rio weight --pause svc=50` or have the equal sign: `rio weight --pause=true svc=50`, so I can change to this:
```
for _, a := range args {
	if a == "pause" || a == "pause=true" {
		paused = true
	}
}
```
But then I'll need to write a new function or manipulate args in order to do rollout. What do you think?

cbron: Helper func 👍

tests/testutil/service.go
rancher-max: Yep exactly! Attach is doing the same thing. I agree we need a logdump though!

tests/validation/weight_test.go
rancher-max: Should be good to go now! Addressed this and the rest of the comments

PR #658, opened 2019-10-31
Update webhook and api types

StrongMonkey:


StrongMonkey:
This PR fixed a bug that service will always be generated even if service is deleted. Also minor bug fixes around github support and PR clean up. This also bump gitwatcher version to fix a log server bug.

StrongMonkey:
https://github.com/rancher/rio/issues

Issue #650, opened 2019-10-31
Add env option to stage command


cbron:
Add this to stage command: https://github.com/rancher/rio/blob/master/cli/cmd/create/create.go#L46
So that a user is able to add and update env variables via the CLI during a stage command. 

This issue is not critical, but a community user has requested it, so would be good to get in sooner.

StrongMonkey:
Available to test in v0.6.0-alpha3.

izaac:
I'm covering both `--env` and `--env-file` here in `rio v0.6.0-alpha.3` 

`--env` is working fine, both `rio export` and connecting to the pod shell show the envrionment variables being set. Same for `--env-file` but with env file it is also copying the comments to the right of the variable value like:

`VARIABLE=true # this is used for blah blah`

All to the right of the equal sign is getting set in the environment variable.

@StrongMonkey @cbron

cbron:
Sounds like its working, closing.

PR #663, opened 2019-10-31
adds env param to stage

cbron:
issue: https://github.com/rancher/rio/issues/650

### DIFF
--- a/cli/cmd/stage/stage.go
+++ b/cli/cmd/stage/stage.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/rancher/rio/pkg/riofile/stringers"
+
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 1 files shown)


cli/cmd/stage/stage.go
```diff
@@ -85,6 +89,11 @@ func (r *Stage) Run(ctx *clicontext.CLIContext) error {
 		if ctx.CLI.String("image") != "" {
 			spec.Image = ctx.CLI.String("image")
 		}
+		stageEnvs, err := stringers.ParseAllEnv(r.EnvFile, r.Env, true)
+		if err != nil {
+			return err
+		}
+		spec.Env = append(spec.Env, stageEnvs...)
```
StrongMonkey: Looks like this will only add new envs but not replacing existing envs

cbron: Hmm, we do need to replace. I'm guessing we want to carry all old ones forward and merge with new ones, but we also want to overwrite any old ones.

Issue #655, opened 2019-10-31
rio -A endpoint Fatal error


rancher-max:
**Describe the bug**
```
rio -A endpoint
FATA[0000] template: :1:2: executing "" at <.Namespace>: can't evaluate field Namespace in type endpoint.Data 
```

**Steps to Reproduce**
1. Run `rio -A endpoint`

**Expected behavior**
Should be able to see endpoints in all of my namespaces without having to do:
`rio endpoint`
`rio -n test-ns endpoint`
... etc

**Rio version**: `rio info`
```
Rio Version: v0.6.0-alpha.2 (d9675e43)
Rio CLI Version: v0.6.0-alpha.2 (d9675e43)
```

StrongMonkey:
Available to test in v0.6.0-alpha3.

rancher-max:
I'm not sure this works yet in v0.6.0-alpha.3. See my example steps below:

1. `rio -n foo run -n demo -p 80 ibuildthecloud/demo:v1`
2. `rio run -n demo -p 80 ibuildthecloud/demo:v3`
3. `rio -A endpoint`

I would expect step 3 to return the endpoints from both "demo" services in different namespaces. Instead, I get this:
```
NAME        ENDPOINTS
dashboard   https://dashboard-rio-system.xxxxxx.on-rio.io:32754
webhook     https://webhook-rio-system.xxxxxx.on-rio.io:32754
demo        https://demo-foo.xxxxxx.on-rio.io:32754
```

davidnuzik:
Ready for testing in alpha4

rancher-max:
Not quite fixed in alpha4. It shows the namespace now, but an overwrite still happens with services of the same name in different namespaces. See example steps above. Output of `rio -A endpoint` looks like this:
```
NAME                    ENDPOINTS
rio-system:webhook      https://webhook-rio-system.xxxxxx.on-rio.io:31358
rio-system:dashboard    https://dashboard-rio-system.xxxxxx.on-rio.io:31358
foo:demo                https://demo-foo.xxxxxx.on-rio.io:31358
```
I would expect one more line: `default:demo      https://demo-default.xxxxxx.on-rio.io:31358`

StrongMonkey:
available to test in v0.6.0-alpha5.

rancher-max:
Fixed in v0.6.0-alpha.5:
```
...
default:demo           https://demo-default.xxxxxx.on-rio.io:<port>
[...]

PR #664, opened 2019-10-31
Fix endpoint cli panic, virtualservices for template

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rio/issues/655

PR #667, opened 2019-11-01
Don't use resource name in permission when run up pod

StrongMonkey:


StrongMonkey:
fix https://github.com/rancher/rio/issues/432

modules/build/controllers/stack/stack.go
```diff
@@ -222,16 +221,6 @@ func populateRbac(stack *riov1.Stack, saName, systemNamespace, buildKitPodName s
...
 
-	role2 := rbac.NewRole(stack.Namespace, fmt.Sprintf("%s-stack", stack.Name), nil)
-	role2.Rules = []v1.PolicyRule{
-		{
-			APIGroups:     []string{"rio.cattle.io"},
-			Resources:     []string{"stacks"},
-			ResourceNames: []string{stack.Name},
-			Verbs:         []string{"get", "update"},
-		},
-	}
```
cbron: This wasn’t used apparently ?

StrongMonkey: Yeah, it is using rio-standard user right now.

PR #669, opened 2019-11-01
Add rio build commands

StrongMonkey:


cli/cmd/builds/builds.go
```diff
@@ -1,22 +1,102 @@
...
+		ls,
+	}
+	return build
+}
+
+type Build struct {
+	T_Tag     string   `desc:"Name and optionally a tag in the 'name:tag' format"`
+	NoCache   bool     `desc:"Do not use cache when building the image"`
+	F_File    string   `desc:"Name of the Dockerfile, defaults to 'Dockerfile'"`
+	BuildArgs []string `desc:"Set build-time variables"`
```
ibuildthecloud:
```suggestion
	BuildArg []string `desc:"Set build-time variables"`
```

cli/cmd/builds/builds.go
```diff
@@ -1,22 +1,102 @@
...
+	P_Push    bool     `desc:"Whether to push images"`
```
ibuildthecloud: Where does it push to?  Explain yourself :)

ibuildthecloud: no push

cli/cmd/images/image.go
```diff
@@ -0,0 +1,123 @@
...
+				for _, tag := range repo.Tags {
+					namespace, name := kv.Split(repo.Name, "/")
+					if namespace != ctx.GetSetNamespace() {
+						continue
+					}
+					result = append(result, Image{
+						Namespace: namespace,
+						Repo:      name,
+						Tag:       tag,
+						Image:     fmt.Sprintf("localhost:5442/%s:%s", repo.Name, tag),
```
cbron: Does this work with external registry ?

cli/cmd/images/image.go
```diff
@@ -0,0 +1,123 @@
...
+	pods, err := ctx.Core.Pods(ctx.SystemNamespace).List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, pod := range pods.Items {
+		if strings.HasPrefix(pod.Name, constants.BuildkitdService) {
+			readyChan := make(chan struct{})
+			go func() {
+				if err := localbuilder.PortForward(ctx.K8s, "9998", "80", pod, false, readyChan, localbuilder.ChanWrapper(ctx.Ctx.Done())); err != nil {
```
cbron: Hardcoding this and URLs below feels kinda gross. Could use constants but that doesn't help all that much. Is there a reason we use both `localhost` and `127.0.0.1` ?

cli/cmd/builds/builds.go
cbron: This is still open

cli/cmd/images/image.go
StrongMonkey: No it doesn't. Those image only lives in private registry.

cli/cmd/images/image.go
StrongMonkey: I will change that to constants. I think for image we always use localhost since it is better for image name. But for url we do url to avoid dns look up.

PR #686, opened 2019-11-05
Fix attach, change webhook name, add rio pods and rio kill

StrongMonkey:


cli/cmd/run/run.go
```diff
@@ -28,9 +30,12 @@ func (r *Run) Run(ctx *clicontext.CLIContext) error {
 		isatty.IsTerminal(os.Stderr.Fd()) &&
 		isatty.IsTerminal(os.Stdin.Fd())
 
+	app, version := services.AppAndVersion(service)
+	svcName := fmt.Sprintf("%s@%s", app, version)
```
cbron: I'm confused after so many changes, are we guaranteeing version now ? Would this never end up as `app@`

cli/main.go
```diff
@@ -196,6 +201,10 @@ func main() {
 			"Apply a rio file",
 			appName+" up [OPTIONS]",
 			""),
+		builder.Command(&kill.Kill{},
+			"Kill pods or pods from services",
```
cbron:
```suggestion
			"Kill pods or services",
```
unless you really wanted to specify that

cli/cmd/run/run.go
StrongMonkey: Yeah this method will always return app and version. The bug in attach is that the parse logic doesn't take the real service name. it only takes app@version.

StrongMonkey:
addressing https://github.com/rancher/rio/issues/691.

Issue #682, opened 2019-11-04
`rio weight` shouldn't change other svc weights


cbron:
Weight without percentage should change spec exactly. Example: If we have 2 services both at 50%, if I call `rio weight svc-b=20`, it should move only svc-b to 20, and not touch svc-a. 

Weight with percentage should change weight relative. Example: If we have 2 services both at 50%, if I call `rio weight svc-b=75%`, then the service should move to 150%. svc-a should not change. 

And fix tests

cbron:
Available to test in v0.6.0-alpha3

izaac:
This is working as described in this issue in the latest `rio v0.6.0-alpha.3`

cbron:
Follow up changes: 

* Weights should be normalized in `rio ps`
* Interval should be normalized
* Interval/Increment should default to 0

izaac:
For the scope of this issue this is passing in `rio v0.6.0-alpha.4`

izaac:
Current behavior is as of `rio v0.6.0-alpha.4`:

> In the backend, the weight command still only edits a single version so it goes up and all others stay as-is. But when you do `rio ps` those should all equal 100%
> 
> The exception to that is when you set any version to 100% with either weight or promote, it will send that version to 100% and all others will get edited back to 0. Because its impossible to be 100% if another version has any weight.

Closing as we're meeting this requirement and this is how weights is behaving at the moment.

PR #695, opened 2019-11-05
Weight cmd calcs percentage on single version

cbron:
Issue: https://github.com/rancher/rio/issues/682


cli/cmd/weight/weight.go
```diff
@@ -78,47 +59,69 @@ func scaleAndAllocate(ctx *clicontext.CLIContext, args []string, pause bool, inc
...
+	serviceName, scaleStr := kv.Split(arg, "=")
+	scaleStr = strings.TrimSuffix(scaleStr, "%")
+	if scaleStr == "" {
+		return serviceName, 0, errors.New("weight params must be in the format of SERVICE=WEIGHT, for example: myservice=10% or myservice=20")
+	}
+	scale, err := strconv.Atoi(scaleStr)
+	if err != nil {
+		return serviceName, scale, fmt.Errorf("failed to parse %s: %v", arg, err)
+	}
+	if scale > 100 {
```
StrongMonkey: if service is a number that exceed 100, looks like error is reported?

cbron: if scale is greater then yes it errors. We don't allow `svc@v2=200` or `svc@v2=200%`. We should try to push people back down to 100 as a total.

PR #702, opened 2019-11-06
Shows user why endpoint is missing

cbron:
Screen Shot 2019-11-06 at 12 34 02 PM


cbron:
Got confirmation from Darren in person.

Issue #660, opened 2019-10-31
Test github workflow


StrongMonkey:
We need to test github workflow. This is done by adding github token to rio. Once token is added, rio will use github webhook to watch push and tag events and creates new services.

We need to test the following scenario: 
1. Rio is able to configure github webhook to github repository.
2. Push and PR event can trigger rio to create a new service.
3. User is able to see deployment status in Github.

StrongMonkey:
@rancher-max don't start testing this until alpha3 is out.

rancher-max:
I used an access token that has all permissions, and by default the webhook rio created is watching `Branch or Tag Creation` and `Pushes`. I believe it should watch `PR` as well by default? When I add that manually to the webhook and create a PR, the deliver fails with response: `{"code":451,"message":"pull request is not enabled","type":"error"}`

Steps I took:
1. `rio secret add --github-webhook` --> keep default namespace and paste my gh access token
2. `rio run -p 8080 -n ghook https://github.com/rancher-max/rio-demo`
3. Validated the webhook created in Github (success, but maybe without all the proper default selections)
4. Create a few changes in master, with PRs and directly committing. In Github, the `Environments` section now looks like this, with everything stuck in Pending even after successful builds:
<img width="914" alt="image" src="https://user-images.githubusercontent.com/55997940/68327748-2b972000-008b-11ea-99ce-a7b6a2f9d139.png">
rio built successfully and updated the service.
5. Added `Pull Requests` to the webhook and submitted a PR. This triggered the hook, but it failed with the error described above.

So overall, I'd say 2 issues with this:
1. PR hooks don't work.
2. User is unable to see deployment statuses in Github.

rancher-max:
Some of that was my fault. This technically fixes both the issues when running with the proper flags: `rio run --build-pr -n gh -p 8080 --template https://github.com/rancher-max/rio-demo`

The only tag that isn't necessary in this example is the `-n` flag for the name. Now this works on PRs and the user is able to see the deployment statuses and endpoints in Github. 

However, it doesn't apply versions, and instead overwrites the service and rolls out weight, which doesn't mean much since they both have the same endpoint now:
```
NAME      IMAGE                                  ENDPOINT                                         PORTS     SCALE     WEIGHT    CREATED              DETAIL
gh        default-gh-n5nqn-ccd1159-pr12:c21d5    https://gh-v0-default.k6yhek.on-rio.io:32754     8080      1         35%       About a minute ago   
[...]

PR #704, opened 2019-11-06
Don't not specify Version for services from template

StrongMonkey:


StrongMonkey:
Address  https://github.com/rancher/rio/issues/660.

PR #705, opened 2019-11-06
Clarify kill command in description

rancher-max:


cli/main.go
```diff
@@ -212,9 +212,9 @@ func main() {
 			appName+" dashboard [OPTIONS]",
 			""),
 		builder.Command(&kill.Kill{},
-			"Kill pods or services",
```
cbron: Given the description I think it would be good to still mention service part. What about: `Kill individual pods, or all pods belonging to a service`.

PR #712, opened 2019-11-07
Fix public domain, endpoint CLI and router

StrongMonkey:


StrongMonkey:
Address https://github.com/rancher/rio/issues/713 https://github.com/rancher/rio/issues/655. Also change the format of `rio router add`. The target can be the regular serviceName([namespace:][type/]app@version)

cli/cmd/route/append.go
```diff
@@ -301,23 +299,21 @@ func convertHeader(data map[string]string) []riov1.HeaderMatch {
...
+func ParseDestinations(ctx *clicontext.CLIContext, targets []string) ([]riov1.WeightedDestination, error) {
 	var result []riov1.WeightedDestination
 	for _, target := range targets {
 		target, optStr := kv.Split(target, ",")
 		opts := kv.SplitMap(optStr, ",")
 
-		// take the last part splitted by /
-		parts := strings.Split(target, "/")
+		_, target = kv.RSplit(target, ":")
+		app, version := kv.Split(target, "@")
```
cbron: `AppAndVersion` ?

StrongMonkey: @cbron target is a string not service

PR #716, opened 2019-11-08
Normalize computed weights and switch to duration

cbron:
Issue: https://github.com/rancher/rio/issues/682


cli/cmd/weight/weight.go
```diff
@@ -3,101 +3,58 @@ package weight
...
 	"github.com/rancher/rio/cli/cmd/util"
 	"github.com/rancher/rio/cli/pkg/clicontext"
+	"github.com/rancher/rio/cli/pkg/table"
+	"github.com/rancher/rio/cli/pkg/types"
 	riov1 "github.com/rancher/rio/pkg/apis/rio.cattle.io/v1"
 	"github.com/rancher/wrangler/pkg/kv"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const DefaultInterval = 2
```
cbron: This could probably be improved upon in future, dynamically scaling for very long or very short durations.

tests/testutil/service.go
```diff
@@ -467,24 +479,25 @@ func (ts *TestService) GetKubeAppEndpointURLs() []string {
...
-	err = json.Unmarshal([]byte(resultString), &svc)
-	if err != nil {
-		ts.T.Fatalf("Failed to unmarshal service results: %s with error: %v", resultString, err.Error())
-	}
-	if svc.Status.ComputedWeight != nil {
-		return *svc.Status.ComputedWeight
-	}
-	return 0
-}
+//func (ts *TestService) GetKubeCurrentWeight() int {
```
cbron: @izaac / @rancher-max I had to comment this out because computedWeight is now relative to all other services (ie: 2 svcs at 100 equals 50% each). This weight percentage is calculated by rio CLI which is what `GetCurrentWeight` now uses. Do we want to do this calc with kubectl too or should we remove ?

izaac: It's ok to comment for now we would need to check if there's a way to get this same value with kubectl, manual testing involved, and validate that against the new CLI calculation.

cli/cmd/weight/weight.go
```diff
@@ -110,11 +67,92 @@ func setPercentageWeight(ctx *clicontext.CLIContext, arg string, rolloutConfig *
...
+	}
+	totalCurrWeightOtherSvcs := totalCurrWeight - currComputedWeight
+
+	newWeight := targetPercentage
+	if targetPercentage == 100 {
+		newWeight = 10000
+	} else if totalCurrWeightOtherSvcs > 0 {
+		// find the weight that would hit our target percentage without touching other service weights
+		// ie: if 2 svcs at 50/50 and you want one at 75%, newWeight would be 150
+		newWeight = int(float64(totalCurrWeightOtherSvcs)/(1-(float64(targetPercentage)/100))) - totalCurrWeightOtherSvcs
```
daxmc99: Revisit in https://github.com/rancher/rio/issues/719

PR #720, opened 2019-11-08
Code cleanup

StrongMonkey:
This PR does the following clean up:
1. remove devMode, move proper arguments to controller
2. move linkerd-job to a stack
3. split rio-controller into controller and  bootstrap stack


StrongMonkey:
Address https://github.com/rancher/rio/issues/721

stacks/bindata-non-static.go
```diff
@@ -0,0 +1,11 @@
+// +build non_static
```
daxmc99: If we are adding build tags that break golangci-lint then we need to also update the golangci-lint configuration

StrongMonkey: @daxmc99 Adding `!static` tag so that by default If you run from IDE it will read localfile instead of bindata

cli/cmd/up/up.go
```diff
@@ -151,7 +151,15 @@ func (u *Up) up(content string, answers map[string]string, s *riov1.Stack, c *cl
...
 	}
 
-	err = c.Apply.WithListerNamespace(c.GetSetNamespace()).WithDefaultNamespace(c.GetSetNamespace()).WithOwner(s).WithSetOwnerReference(true, true).WithGVK(knowngvks...).WithDynamicLookup().ApplyObjects(objs...)
+	err = c.Apply.
+		WithListerNamespace(c.GetSetNamespace()).
+		WithDefaultNamespace(c.GetSetNamespace()).
+		WithOwner(s).
+		WithSetOwnerReference(true, true).
+		WithGVK(knowngvks...).
+		WithRestrictClusterScoped().
```
daxmc99: 👍  This means riofiles cannot have cluster scoped resources correct?

StrongMonkey: correct

Issue #641, opened 2019-10-30
Handle rio export


cbron:
Stack export should fetch objects, like namespace export does.:  https://github.com/rancher/rio/blob/ac038a30ec1c2c6a8e373b2b495d4815c88e1773/cli/cmd/export/export.go#L49-L51
Will likely need to update the riofile integration test. 
 
Also need a solution for edit: https://github.com/rancher/rio/blob/master/cli/cmd/edit/stack/stack.go

Finally, fix riofile export tests

davidnuzik:
Possibly impacts other areas that may utilize template field.

cbron:
If edit is non-trivial to update (ie we can't just reuse the export logic and call `up` again) bring up idea to wider group about removing stack edit and having people use the manual `export -> edit -> up` flow.

cbron:
Ready to test in v0.6.0-alpha.5, ping me for questions

izaac:
I've been trying to do a `rio up` with the default demo but also on @daxmc99's Riofile with k8s manifest but I'm not able to get the stack created somehow in alpha.5 @StrongMonkey :

I've tried testing this using the default Kubeconfig retrieved from a k3s cluster (copied from its master node) and a GKE cluster (using gcloud) without any RBAC users configuration.

`rio up --name teststack --branch track_gvk  https://github.com/daxmc99/rio-demo`

`rio logs <taskrun>` shows:

```
default-teststack-stack-629c5-dfbd1-pod-f00d95 step-rio-up time="2019-11-12T04:00:40Z" level=fatal msg="failed to create default/frontend apps/v1, Kind=Deployment for  default/teststack: deployments.apps is forbidden: User \"system:serviceaccount:default:default-teststack-stack-629c5-dfbd1-stack\" cannot create resource \"deployments\" in API group \"apps\" in the namespace \"default\", failed to create default/redis-master apps/v1, Kind=Deployment for  default/teststack: deployments.apps is forbidden: User \"system:serviceaccount:default:default-teststack-stack-629c5-dfbd1-stack\" cannot create resource \"deployments\" in API group \"apps\" in the namespace \"default\", failed to create default/redis-slave apps/v1, Kind=Deployment for  default/teststack: deployments.apps is forbidden: User \"system:serviceaccount:default:default-teststack-stack-629c5-dfbd1-stack\" cannot create resource \"deployments\" in API group \"apps\" in the namespace \"default\""
```

This other using `--permissions` flags too but it failed the RBAC role:

[...]

PR #722, opened 2019-11-08
Enable export of stack by gvk

cbron:
Issue: https://github.com/rancher/rio/issues/641

### DIFF
--- a/cli/cmd/export/export.go
+++ b/cli/cmd/export/export.go
@@ -10,10 +10,12 @@ import (
 	riov1 "github.com/rancher/rio/pkg/apis/rio.cattle.io/v1"
 	"github.com/rancher/rio/pkg/riofile"
 	"github.com/rancher/wrangler/pkg/gvk"
+	name "github.com/rancher/wrangler/pkg/name"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"sigs.k8s.io/yaml"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 )
 
 type Export struct {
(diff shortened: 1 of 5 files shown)


cli/cmd/export/export.go
```diff
@@ -82,6 +93,60 @@ func (e *Export) Run(ctx *clicontext.CLIContext) error {
...
+					return nil, err
+				}
+				objects = append(objects, es)
+			} else if item.GetKind() == "Router" {
+				r := &riov1.Router{}
+				if err := json.Unmarshal(data, r); err != nil {
+					return nil, err
+				}
+				objects = append(objects, r)
+			}
```
StrongMonkey: @cbron Do you export any object that is not Rio types? like any k8s manifest you can define in Riofile.

cli/cmd/export/export.go
```diff
@@ -82,6 +93,60 @@ func (e *Export) Run(ctx *clicontext.CLIContext) error {
...
+	var objects []runtime.Object
+	client, err := dynamic.NewForConfig(ctx.Config.RestConfig)
+	if err != nil {
+		return nil, err
+	}
+	for _, gvk := range stack.Spec.AdditionalGroupVersionKinds {
+		gvr := schema.GroupVersionResource{
+			Group:    gvk.Group,
+			Version:  gvk.Version,
+			Resource: strings.ToLower(name.GuessPluralName(gvk.Kind)),
```
daxmc99: Do we need to fix this where we ingest gvks to ensure they have the plural name?

cli/cmd/export/export.go
```diff
@@ -82,6 +93,60 @@ func (e *Export) Run(ctx *clicontext.CLIContext) error {
...
+		}
+		resourceClient := client.Resource(gvr)
+		ul, err := resourceClient.List(metav1.ListOptions{
```
daxmc99: You should be able to list only resources in the current namespace since we scope stacks to namespaces

cli/cmd/export/export.go
cbron: fixed

Issue #709, opened 2019-11-07
Public domain not returning the proper service content, http returns 404 [v0.6.0-alpha.3]


izaac:
**Describe the bug**
Registering a domain to a service is working but the response is wrongly returning 404 and secure http just halts the connection.

**To Reproduce**
1. Use a CNAME with the rio cluster domain as value
2. register this public domain (CNAME'd) and associate to a service that has a valid ENDPOINT.
3. Curl the public domain (CNAME) http is responding 404 and secure http is closing connection

**Expected behavior**

I should get the service content equals to what it's endpoint is returning but using the publicdomain

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.6", GitCommit:"96fac5cd13a5dc064f7d9f4f23030a6aeface6cc", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:49Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.6-eks-5047ed", GitCommit:"5047edce664593832e9b889e447ac75ab104f527", GitTreeState:"clean", BuildDate:"2019-08-21T22:32:40Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.6.0-alpha.3 (98e09134)
Rio CLI Version: v0.6.0-alpha.3 (98e09134)
Cluster Domain: xxxxx.on-rio.io
Cluster Domain IPs:
System Namespace: rio-system
Wildcard certificates: true
```
**Additional context**
`rio systemlogs`
output:
```
No related information found in systemlogs.
```

izaac:
Related to the fix coming for `rio v0.6.0-alpha.4` https://github.com/rancher/rio/pull/703

StrongMonkey:
@izaac Available to test in alpha4

izaac:
This is working in `rio v0.6.0-alpha.4` I've created a CNAME and assigned to the Cluster Domain. 
Then I've registered the CNAME as a public domain with rio associating it to a current service. 

But after adding a few domains I got an error when I request HTTPS traffic:

```
[...]

PR #725, opened 2019-11-09
Fix public domain bug round 3

StrongMonkey:
Fix regression bug that were introduced in api refactoring


StrongMonkey:
addressing https://github.com/rancher/rio/issues/709.

PR #729, opened 2019-11-11
Fix adding docker secret in CLI

StrongMonkey:


StrongMonkey:
No test since CLI is based on live prompt. We need to come with a better way to test this.

Issue #727, opened 2019-11-09
rio up not printing the created stack name to stdout [v0.6.0-alpha.4]


izaac:
**Describe the bug**
Performing a `rio up -f [path to Riofile]` or `rio up [github url with Riofile` isn't printing the stack name created when completed successfully.

**To Reproduce**
1. rio up -f [path to Riofile]
2. wait until it completes applying the Riofile.
3. No output is shown

**Expected behavior**

rio should print the created stack name when successfully applying a Riofile.
`rio stack ls` should show the same.

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.6", GitCommit:"96fac5cd13a5dc064f7d9f4f23030a6aeface6cc", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:49Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2-k3s.1", GitCommit:"b8b17ba55f20e590df507fce333dfee13ab438c6", GitTreeState:"clean", BuildDate:"2019-10-16T05:17Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.6.0-alpha.4 (969baf29)
Rio CLI Version: v0.6.0-alpha.4 (969baf29)
Cluster Domain: xxxxxx.on-rio.io
Cluster Domain IPs: x.x.x.x
System Namespace: rio-system
Wildcard certificates: true
```
**Additional context**
`rio systemlogs`
output:
```
No related systemlogs found.
```

StrongMonkey:
available to test in v0.6.0-alpha5.

izaac:
Confirmed working in `rio v0.6.0-alpha5`

PR #731, opened 2019-11-11
Fix stack name not printed when doing up

StrongMonkey:


StrongMonkey:
fix https://github.com/rancher/rio/issues/727.

Issue #736, opened 2019-11-11
Running service with --build-pr flag promotes PR versions immediately


rancher-max:
**Describe the bug**
https://github.com/rancher/rio/issues/660#issuecomment-552540314

**To Reproduce**
1. `rio secret add --github-webhook` --> keep default namespace and paste my gh access token
2. `rio run --build-pr -n gh -p 8080 --template https://github.com/rancher-max/rio-demo` --> Validated the webhook created in Github
3. Create a new branch, make a change to the code in the branch, and then submit a PR to master.

This created a deployment to `staging` environment in Github, but in rio automatically promoted that service to 100% weight, so it effectively overwrote production as well.

**Expected behavior**
New revisions from PRs (setting the `--build-pr` flag with `rio run`) should not automatically promote revisions created from the PR.

**Rio version**: `v0.6.0-alpha.4`

StrongMonkey:
available to test in v0.6.0-alpha5.

rancher-max:
v0.6.0-alpha.5 works great. This is my favorite flow now :)

PR #737, opened 2019-11-11
Fix staging bug when specifying template and PR

StrongMonkey:


StrongMonkey:
Addressing https://github.com/rancher/rio/issues/736.

Issue #739, opened 2019-11-11
[VISUAL ONLY] rio up with multiple services without apps specified aggregates them all to one


rancher-max:
**Describe the bug**
If the `app` is not explicitly specified in a riofile, it is aggregating all of the `app`less services together.

**To Reproduce**
1. `rio up https://github.com/rancher-max/riofile-demo`
2. `rio ps`
 --> Notice how the weights for three of the services without apps are showing all at 33%. They should all be separately 100%. Their endpoints show properly and seem to work as expected.

**Expected behavior**
Services without apps should not aggregate together. They have different `.status.computedApp` values, and this should be taken into account as their "app"

Type:
**Rio version**: v0.6.0-alpha.4

rancher-max:
Fixed in v0.6.0-rc1

PR #741, opened 2019-11-12
Show svc weight as 100 if app blank

cbron:
Issue: https://github.com/rancher/rio/issues/739

Weight calc on ps was summing on app string, but app string can be blank and it was counting those as the same app.

### DIFF
--- a/cli/cmd/ps/ps_services.go
+++ b/cli/cmd/ps/ps_services.go
@@ -9,6 +9,7 @@ import (
 	"github.com/rancher/rio/cli/pkg/tables"
 	clitypes "github.com/rancher/rio/cli/pkg/types"
 	riov1 "github.com/rancher/rio/pkg/apis/rio.cattle.io/v1"
+	"github.com/rancher/rio/pkg/services"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


cli/cmd/ps/ps_services.go
```diff
@@ -121,13 +121,17 @@ func (p *Ps) services(ctx *clicontext.CLIContext) error {
 
 		weight := 0
 		if service.(*riov1.Service).Status.ComputedWeight != nil && *service.(*riov1.Service).Status.ComputedWeight > 0 {
-			totalWeight := 0
-			for _, subService := range services {
-				if service.(*riov1.Service).Spec.App == subService.(*riov1.Service).Spec.App && subService.(*riov1.Service).Status.ComputedWeight != nil {
-					totalWeight += *subService.(*riov1.Service).Status.ComputedWeight
+			if service.(*riov1.Service).Spec.App == "" {
```
StrongMonkey: @cbron You should always use `appAndVersion` method to get app.

cbron: Fixed

cli/cmd/ps/ps_services.go
```diff
@@ -114,20 +115,26 @@ func (p *Ps) services(ctx *clicontext.CLIContext) error {
...
 			continue
 		}
 
 		weight := 0
 		if service.(*riov1.Service).Status.ComputedWeight != nil && *service.(*riov1.Service).Status.ComputedWeight > 0 {
-			totalWeight := 0
-			for _, subService := range services {
-				if service.(*riov1.Service).Spec.App == subService.(*riov1.Service).Spec.App && subService.(*riov1.Service).Status.ComputedWeight != nil {
-					totalWeight += *subService.(*riov1.Service).Status.ComputedWeight
+			app, _ := services.AppAndVersion(service.(*riov1.Service))
```
StrongMonkey: Hmm, app will not be empty in this case.

cbron: Ohh, I bet its just falling down to next check and that check works correctly now. I'll remove.

Issue #713, opened 2019-11-07
`rio endpoint` doesn't show publicdoman as https


StrongMonkey:
Steps to reproduce:

1. Register a domain. `rio domain register`
2. `rio endpoints`
3. Publicdomain endpoints are not showed as `https`

StrongMonkey:
available to test in v0.6.0-alpha5.

rancher-max:
On GKE this gave the http edpoint after `rio endpoint` but on k3s it gave https endpoint. I believe @StrongMonkey has a fix incoming for this.

rancher-max:
Appears to be fixed in the cluster that was showing http endpoint previously with rio version v0.6.0-rc1

PR #746, opened 2019-11-12
Fix indexer on the wrong namespace

StrongMonkey:


StrongMonkey:
fix https://github.com/rancher/rio/issues/713.

PR #755, opened 2019-11-12
Add some simple retry to service rio commands

rancher-max:
Hopefully this helps with flakiness...

### DIFF
--- a/tests/testutil/service.go
+++ b/tests/testutil/service.go
@@ -32,7 +32,7 @@ type TestService struct {
 // returns a new TestService with it attached. Guarantees ready state but not live endpoint
 func (ts *TestService) Create(t *testing.T, source ...string) {
 	args, envs := ts.createArgs(t, source...)
-	_, err := RioCmd(args, envs...)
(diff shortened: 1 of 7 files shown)


cbron:
Feel like this should at least be wrapped in a helper.

rancher-max:
> Feel like this should at least be wrapped in a helper.

Yeah it probably should. If it works I will update

tests/testutil/service.go
```diff
@@ -137,7 +137,13 @@ func (ts *TestService) Attach() []string {
 
 // Call "rio scale ns/service={scaleTo}"
 func (ts *TestService) Scale(scaleTo int) {
-	_, err := RioCmd([]string{"scale", fmt.Sprintf("%s=%d", ts.Name, scaleTo)})
+	var err error
+	for i := 0; i < 5; i++ {
```
StrongMonkey: Also doing some backoff sleep would be better

rancher-max:
Going to retry build a few times just to be sure this actually fixes the flaky tests, then we can merge

PR #757, opened 2019-11-12
alias logs cmd

cbron:


cli/cmd/logs/logs.go
```diff
@@ -31,6 +32,10 @@ type Logs struct {
 
 // This is based on both wercker/stern and linkerd/stern implementations
 
+func (l *Logs) Customize(cmd *cli.Command) {
+	cmd.ShortName = "log"
```
StrongMonkey: Wait, shouldn't that be alias

cbron: https://github.com/rancher/rio/blob/master/cli/cmd/images/image.go#L47-L49

StrongMonkey: ahhh
> // short name of the command. Typically one character (deprecated, use `Aliases`)

StrongMonkey: I was wrong for the usage, but should be fine.

Issue #758, opened 2019-11-12
Setting --weight flag when running new services fails


rancher-max:
**Describe the bug**
`FATA[0000] no services found` when setting the `--weight` flag on `rio run`

**To Reproduce**
1. `rio run --weight=100 -p 80 ibuildthecloud/demo:v1`

**Expected behavior**
It should create the service with a spec weight of 100

Type:
**Rio version**: `rio version v0.6.0-alpha.5 (ac4bbfee)`

rancher-max:
#755 stole the change from #759

rancher-max:
Fixed in v0.6.0-rc1

PR #759, opened 2019-11-13
Allows spec weight to be set on first app version

cbron:
issue: https://github.com/rancher/rio/issues/758


cbron:
Closing, already implemented

Issue #763, opened 2019-11-13
Installing rio from YAML fails


dnoland1:
**Describe the bug**
When generating the rio yaml and attempting to apply it, it fails to create the deployment since the namespace is not created.

**To Reproduce**
```
$ rio install --yaml > rio.yaml
$ kubectl apply -f rio.yaml
namespace/rio-system created
clusterrole.rbac.authorization.k8s.io/rio-cluster-admin created
clusterrolebinding.rbac.authorization.k8s.io/rio-controller-binding created
serviceaccount/rio-controller-serviceaccount created
secret/rio-api-validator created
service/rio-api-validator created
clusterrole.rbac.authorization.k8s.io/rio-admin created
clusterrole.rbac.authorization.k8s.io/rio-readonly created
clusterrole.rbac.authorization.k8s.io/rio-privileged created
clusterrole.rbac.authorization.k8s.io/rio-standard created
configmap/rio-config created
Error from server (NotFound): error when creating "rio.yaml": namespaces "rio-system" not found
```

**Expected behavior**
No error and it should work. The namespace yaml should be before the deployment yaml.

Workaround, run kubectl apply a second time.

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:41:55Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.2", GitCommit:"f6278300bebbb750328ac16ee6dd3aa7d3549568", GitTreeState:"clean", BuildDate:"2019-08-05T09:15:22Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
$ rio info
Rio Version: v0.6.0-rc1 (1dec45ce)
Rio CLI Version: v0.6.0-rc1 (1dec45ce)
Cluster Domain: cnphoc.on-rio.io
Cluster Domain IPs: 10.0.2.10
System Namespace: rio-system
Wildcard certificates: true
```
**Additional context**
`rio systemlogs`
output:
```

```

izaac:
default namespace yaml generation and kubectl apply worked.

```
[...]

PR #764, opened 2019-11-13
Fix yaml output order

StrongMonkey:


StrongMonkey:
fix https://github.com/rancher/rio/issues/763

PR #766, opened 2019-11-13
Fix pvc informer

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rio/issues/767.

StrongMonkey:
@cbron I added one commit that seems to solve one of the race condition error during bootstrapping. Now smi feature will create CRD first before watching traffic.split.

cbron:
Looks like install never finished ?

Issue #750, opened 2019-11-12
health-url param on rio run errors out


StrongMonkey:
`rio run --health-url http://locallost:80 nginx` doesn't work
Throw an validation error if you run it.

StrongMonkey:
Actually appears to be a server side error

StrongMonkey:
We should fix this v0.6.0 release.

izaac:
`rio v0.6.0-rc1` and below versions error out with:

```
rio run --health-url http://localhost:80 -p 80 nginx
FATA[0000] Service.rio.cattle.io "modest-tu-5brlz" is invalid: []: Invalid value: map[string]interface {}{"apiVersion":"rio.cattle.io/v1", "kind":"Service", "metadata":map[string]interface {}{"creationTimestamp":"2019-11-14T21:48:29Z", "generateName":"modest-tu-", "generation":1, "name":"modest-tu-5brlz", "namespace":"default", "uid":"7eb1c50c-0728-11ea-88c9-42010aa80038"}, "spec":map[string]interface {}{"app":"modest-tu", "image":"nginx", "ports":[]interface {}{map[string]interface {}{"port":80, "targetPort":80}}, "readinessProbe":map[string]interface {}{"httpGet":map[string]interface {}{"host":"localhost:80", "port":80, "scheme":"HTTP"}}, "version":"v0"}}: validation failure list:
spec.readinessProbe.httpGet.port in body must be of type string: "integer"
```

Now the `--health-url` flag is properly setting the Readiness Check in the service:

```yaml
   spec:
      automountServiceAccountToken: true
      containers:
      - image: nginx
        imagePullPolicy: Always
        name: epic-diffie-v0hrlzp
        ports:
        - containerPort: 80
          protocol: TCP
        readinessProbe:
          failureThreshold: 3
          httpGet:
            host: localhost
            path: /
            port: 80
            scheme: HTTP
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 1
        resources:
          requests:
            cpu: 50m
            memory: 64Mi
```

```
Rio Version: a3e458ef (a3e458ef)
Rio CLI Version: dev (HEAD)
Cluster Domain: xxxxxx.on-rio.io
Cluster Domain IPs: x.x.x.x
System Namespace: rio-system
Wildcard certificates: true
```

PR #770, opened 2019-11-13
Fixes health-url cli and backend, improves examples

cbron:
Issue: https://github.com/rancher/rio/issues/750

Examples: 
* `rio run --health-url http://:80 -p 80 nginx`
* `rio run --health-url http://:8080/status --health-initial-delay 10s --health-interval 5s --health-failure-threshold 5 --health-timeout 5s -p 8080 cbron/mybusybox:dev`

### DIFF
--- a/cli/cmd/create/create.go
+++ b/cli/cmd/create/create.go
@@ -56,7 +56,7 @@ type Create struct {
 	HealthInterval         string            `desc:"Time between running the check (ms|s|m|h)" default:"0s"`
 	HealthSuccessThreshold int               `desc:"Consecutive successes needed to report healthy"`
 	HealthTimeout          string            `desc:"Maximum time to allow one check to run (ms|s|m|h)" default:"0s"`
-	HealthURL              string            `desc:"URL to hit to check health (example: http://localhost:8080/ping)"`
+	HealthURL              string            `desc:"URL to hit to check health (example: http://:8080/ping)"`
 	Hostname               string            `desc:"Container host name"`
 	I_Interactive          bool              `desc:"Keep STDIN open even if not attached"`
 	ImagePullPolicy        string            `desc:"Behavior determining when to pull the image (never|always|not-present)" default:"not-present"`
--- a/cli/cmd/create/health.go
+++ b/cli/cmd/create/health.go
@@ -3,7 +3,6 @@ package create
 import (
 	"fmt"
 	"net/url"
-	"strconv"
 	"time"
 
 	"github.com/mattn/go-shellwords"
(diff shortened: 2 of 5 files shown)


cli/cmd/create/health.go
```diff
@@ -69,19 +68,14 @@ func (c *Create) setHealthCheck(spec *riov1.ServiceSpec) error {
...
 			hc.TCPSocket = &v1.TCPSocketAction{
-				Port: intstr.FromInt(port),
+				Port: intstr.FromString(portStr),
 			}
 		} else {
 			hc.HTTPGet = &v1.HTTPGetAction{
-				Port: intstr.FromInt(port),
-				Host: u.Host,
+				Port: intstr.FromString(portStr),
+				Host: u.Hostname(),
```
cbron: `Host` includes port which doubles it, `Hostname()` removes it

cli/cmd/create/create.go
```diff
@@ -56,7 +56,7 @@ type Create struct {
 	HealthInterval         string            `desc:"Time between running the check (ms|s|m|h)" default:"0s"`
 	HealthSuccessThreshold int               `desc:"Consecutive successes needed to report healthy"`
 	HealthTimeout          string            `desc:"Maximum time to allow one check to run (ms|s|m|h)" default:"0s"`
-	HealthURL              string            `desc:"URL to hit to check health (example: http://localhost:8080/ping)"`
+	HealthURL              string            `desc:"URL to hit to check health (example: http://:8080/ping)"`
```
StrongMonkey: why this change?

modules/service/controllers/service/populate/pod/containers.go
```diff
@@ -87,6 +87,17 @@ func toContainer(containerName string, c *riov1.Container) v1.Container {
...
+		// Norman only takes string for intstr types, but K8s only takes int, so do conversion
+		if c.ReadinessProbe.HTTPGet != nil && c.ReadinessProbe.HTTPGet.Port.StrVal != "" {
+			c.ReadinessProbe.HTTPGet.Port = intstr.FromInt(c.ReadinessProbe.HTTPGet.Port.IntValue())
+		}
+		if c.ReadinessProbe.TCPSocket != nil && c.ReadinessProbe.TCPSocket.Port.StrVal != "" {
+			c.ReadinessProbe.TCPSocket.Port = intstr.FromInt(c.ReadinessProbe.TCPSocket.Port.IntValue())
+		}
+		con.ReadinessProbe = c.ReadinessProbe
+	}
+
```
StrongMonkey: Same for liveness?

cli/cmd/create/create.go
cbron: It's misleading to include in example, localhost won't work for most cases:
```The kubelet sends the probe to the pod’s IP address, unless the address is overridden by the optional host field in httpGet. If scheme field is set to HTTPS, the kubelet sends an HTTPS request skipping the certificate verification. In most scenarios, you do not want to set the host field. ```

Issue #775, opened 2019-11-13
External IP not created properly with externalservice from other namespace


rancher-max:
**Describe the bug**
Creating an externalservice pointing to a service in a different namespace does not properly create the external-ip

**To Reproduce**
1. Create a service in a different namespace: `rio -n foo run -p 80 -n what ibuildthecloud/demo:v1`
2. Create an externalservice in the default namespace to point to that service: `rio external create foowhat foo:what`
3. `kubectl get svc -A` and the `foowhat` service should have the expected External IP. Currently it is `what-v0.foo`

**Expected behavior**
External-IP should be like `what-v0.foo.svc.cluster.local`

rancher-max:
This is fixed in v0.6.0-rc2. Also just to verify, executing `curl foowhat` from inside a container of a pod in the default namespace does give the valid response (`Hello World`)

PR #776, opened 2019-11-14
Fix external service bug

StrongMonkey:
related issue: https://github.com/rancher/rio/issues/775

### DIFF
--- a/cli/cmd/externalservice/create.go
+++ b/cli/cmd/externalservice/create.go
@@ -38,7 +38,9 @@ func (c *Create) Run(ctx *clicontext.CLIContext) error {
 				externalService.Spec.TargetServiceNamespace = ref.Namespace
 			} else {
 				externalService.Spec.TargetApp = ref.App
-				externalService.Spec.TargetVersion = ref.Version
+				if strings.Contains(name, "@") {
+					externalService.Spec.TargetVersion = ref.Version
+				}
 				externalService.Spec.TargetServiceNamespace = ref.Namespace
 			}
 		}
--- a/modules/service/controllers/externalservice/populate/populate.go
+++ b/modules/service/controllers/externalservice/populate/populate.go
@@ -57,7 +57,7 @@ func getSpec(es *riov1.ExternalService, os *objectset.ObjectSet) (v1.ServiceSpec
 		}
 		return v1.ServiceSpec{
 			Type:         v1.ServiceTypeExternalName,
-			ExternalName: fmt.Sprintf("%s.%s", service, ns),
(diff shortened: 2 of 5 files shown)


cli/cmd/externalservice/create.go
```diff
@@ -38,7 +38,9 @@ func (c *Create) Run(ctx *clicontext.CLIContext) error {
 				externalService.Spec.TargetServiceNamespace = ref.Namespace
 			} else {
 				externalService.Spec.TargetApp = ref.App
-				externalService.Spec.TargetVersion = ref.Version
+				if strings.Contains(name, "@") {
```
cbron: I don't understand this, `ParseID` returns the version, what does `@` sign signify ?

modules/service/controllers/externalservice/populate/populate.go
```diff
@@ -57,7 +57,7 @@ func getSpec(es *riov1.ExternalService, os *objectset.ObjectSet) (v1.ServiceSpec
 		}
 		return v1.ServiceSpec{
 			Type:         v1.ServiceTypeExternalName,
-			ExternalName: fmt.Sprintf("%s.%s", service, ns),
+			ExternalName: fmt.Sprintf("%s.%s.svc.cluster.local", service, ns),
```
cbron: Without this it doesn't scope to the cluster ?

tests/integration/externalservice_test.go
```diff
@@ -31,4 +31,12 @@ func externalServiceTests(t *testing.T, when spec.G, it spec.S) {
 			assert.Equal(t, externalService.GetFQDN(), externalService.GetKubeFQDN())
 		})
 	}, spec.Parallel())
+	when("a config is created pointing to another service", func() {
```
cbron:
```suggestion
	when("an externalservice is created pointing to another service", func() {
```

cli/cmd/externalservice/create.go
StrongMonkey: if format is `rio external create ext app`  then externalservice points to `app`

if format is `rio external create ext app@version` then externalservice points to `app` and `version`.

modules/service/controllers/externalservice/populate/populate.go
StrongMonkey: yes, it is a cname record so it will return $name.$namespace. Kubernetes use search path to resolve that but looks like it is not getting resolved for cname record.

cli/cmd/externalservice/create.go
cbron: Makes sense. A helper method for app vs app/version might be nice and more obvious at first glance. But fine for now.

PR #777, opened 2019-11-14
Use generated names

StrongMonkey:


StrongMonkey:
In CLI we should always set the GenerateName, This matches what UI does.

Issue #783, opened 2019-11-14
Warn on install on any K8s version under 1.15


cbron:
- Concerned about parsing
- Could warn api validation isn't done on xyz versions

izaac:
Warning is showing for the following kubernetes versions and only shown once after `rio install`:
All the clusters have more than one worker.

1.13.x and 1.14.x

```
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.11-gke.14", GitCommit:"56d89863d1033f9668ddd6e1c1aea81cd846ef88", GitTreeState:"clean", BuildDate:"2019-11-07T19:12:22Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}

Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.8-gke.12", GitCommit:"188432a69210ca32cafded81b4dd1c063720cac0", GitTreeState:"clean", BuildDate:"2019-11-07T19:27:01Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}

```

Warning was not showing during testing on the following k8s versions:


1.15.x and 1.16.x
```
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.4-gke.18", GitCommit:"efb62569d2358ca32843f553f190c2456878e253", GitTreeState:"clean", BuildDate:"2019-10-21T20:05:39Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}

Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```

```
Rio Version: a4fc0e0f-dirty (a4fc0e0f)
Rio CLI Version: dev (HEAD)
Cluster Domain: xxxxxx.on-rio.io
Cluster Domain IPs: x.x.x.x
System Namespace: rio-system
Wildcard certificates: false
```

izaac:
With one worker the warning is also shown once:

```
[...]

PR #786, opened 2019-11-14
Warn on unsupported kubernetes versions

cbron:
issue: https://github.com/rancher/rio/issues/783

### DIFF
--- a/cli/cmd/install/install.go
+++ b/cli/cmd/install/install.go
@@ -7,8 +7,7 @@ import (
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cli/cmd/install/install.go
```diff
@@ -175,6 +182,17 @@ func (i *Install) preConfigure(ctx *clicontext.CLIContext, ignoreCluster bool) e
 	return nil
 }
 
+func checkKubernetesVersion(version string) {
```
StrongMonkey: hmm, can we use semver to do the check?

PR #787, opened 2019-11-14
Update gloo vendor

StrongMonkey:


StrongMonkey:
Not going to do that for this upcoming release

PR #788, opened 2019-11-14
only display k8s version warning once

cbron:
issue: https://github.com/rancher/rio/issues/783

### DIFF
--- a/cli/cmd/install/install.go
+++ b/cli/cmd/install/install.go
@@ -163,9 +163,12 @@ func (i *Install) preConfigure(ctx *clicontext.CLIContext, ignoreCluster bool) e
 	}
 
 	var totalMemory int64
+	versionWarning := false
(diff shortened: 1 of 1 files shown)


cli/cmd/install/install.go
```diff
@@ -163,9 +163,12 @@ func (i *Install) preConfigure(ctx *clicontext.CLIContext, ignoreCluster bool) e
 	}
 
 	var totalMemory int64
+	versionWarning := false
 	for _, node := range nodes.Items {
 		totalMemory += node.Status.Capacity.Memory().Value()
-		checkKubernetesVersion(node.Status.NodeInfo.KubeletVersion)
```
StrongMonkey: LGTM, can be simplied to save the version first. Then compare version outside the loop. Minor detail.

Issue #511, opened 2019-09-25
Update Docs and Wiki for 0.6.0


cbron:
Updates 

* api changes, like removing app, image builder
* workflow changes in riofile, and ensure other example flows still work
* ...

- [x] index
- [x] autoscaling
- [x] cli-reference
- [x] concept
- [x] CD
- [x] dashboard
- [x] dev-app
- [x] faq
- [x] install
- [x] monitoring
- [x] publicdomain-external
- [x] quick-stack
- [x] rbac
- [x] riofile
- [x] router
- [x] workload
- [x] Move 0.6.0 to primary docs
- [x] Make docs more prominent on home readme

Address comments

- [x] ports in run examples
- [x] rio stage shows new options
- [x] rio rm can be run with multiple services
- [x] no-deploy traefik
- [x] initial version is v0
- [x] expected weight behavior

davidnuzik:
This issue captures the minimum changes we need to update docs for v0.6.0

cbron:
Ensure port is in all run examples, as that is required now.

rancher-max:
Ensure `rio stage` shows new options for environment variables and calls out `--edit` caveats: https://github.com/rancher/rio/issues/673#issuecomment-550422346

rancher-max:
Document that `rio rm` can be run with multiple services, e.g: `rio rm demo stack/test othersvc`

izaac:
`--no-deploy traefik` Ensure that option is recommended for k3s cluster installs. There are known issues with traefik and rio. Most of them related to Endpoints/PublicDomains.

rancher-max:
Document somewhere that the initial version is created with `@v0`. This could be useful if someone wants to specifically create a route or publicdomain to that service version instead of the "app"

izaac:
Be clear in docs on the weights behavior: For example this was expected:
https://github.com/rancher/rio/issues/769

PR #810, opened 2019-11-15
Document some FAQs

rancher-max:


cbron:
Should we add port endpoint thing ? Its more obvious now with the message but...

rancher-max:
> Should we add port endpoint thing ? Its more obvious now with the message but...

I was thinking that might be useful too. And pretty much every comment from here: https://github.com/rancher/rio/issues/511

PR #821, opened 2019-11-18
Add Riofile example, fix minor doc bugs

StrongMonkey:


docs/v0.6.0/webhooks.md
```diff
@@ -18,3 +18,14 @@ rio run -p 80 --build-webhook-secret=githubtoken https://github.com/example/exam
...
 
+# Webhook for Riofile
+
+Setup webhook and private git clone secret for git repository that contains Riofile:
+
+```bash
+$ rio up --build-clone-secret gitsecret --build-webhook-secret webhook https://github.com/example/example
+```
+
+For how to add git and webhook secret, check [here](./continuous-deployment.md).
```
cbron: Should this be `webhooks.md` ?

docs/v0.6.0/riofile.md
```diff
@@ -526,3 +523,62 @@ guestbook   https://guestbook-default.enu90s.on-rio.io
 We can now access this endpoint over encrypted https!
 
 
+#### Using answer file
```
cbron: Good stuff 👍 . There is a section above this that shows small template usage but this is better.

docs/v0.6.0/webhooks.md
StrongMonkey: The docs for adding secrets actually lives in continuous-deployment.md. I think we can keep it there.

PR #825, opened 2019-11-18
Fix volume mapper

StrongMonkey:


StrongMonkey:
Address https://github.com/rancher/rio/issues/826.

Issue #774, opened 2019-11-13
`rio run` with weight 100 doesn't de-promote other services


cbron:
`rio run -n z -p 80 nginx`
`rio run --weight 100 -n z@v2 -p 80 nginx`
Result: v2 only gets 99% weight

rancher-max:
Working in master now. Weight behavior is looking great!

Issue #719, opened 2019-11-08
Verify weight calculations and add unit tests 


daxmc99:
The weight calculation in the rollout controller and weight/promote CLI needs to more formally verified to address possible edge cases. 

- [x] duration may need to scale dynamically up and down instead of being stuck at 2 seconds
- [x] (from comments) look into making rollout more gradual, especially with larger durations
- [x] math can be broken out and unit tested
- [x] check for rounding (floor/ceiling) issues, especially around weight setting

daxmc99:
Also need to test for high conflict levels during rollout per @izaac

rancher-max:
Just something to review as part of this:
`rio weight --duration=1m test@v2=60` Seems like it should rollout 1% per second, but does 20% in the first 10 seconds, is pretty gradual in the middle, then takes around 25 seconds for the final 10%.
This is somewhat expected as part of the "only-touch-one-service" system, but I think should just be double checked into our options here.

izaac:
> Just something to review as part of this:
> `rio weight --duration=1m test@v2=60` Seems like it should rollout 1% per second, but does 20% in the first 10 seconds, is pretty gradual in the middle, then takes around 25 seconds for the final 10%.
> This is somewhat expected as part of the "only-touch-one-service" system, but I think should just be double checked into our options here.

This is even more noticeable when using hours.

rancher-max:
Weight is all an approximation, but this is working much much better now in master. It is pretty gradual now.

I would recommend users who want to slowly rollout weight, but overall aren't worried about their new version stability, to use the gradual duration feature.
I would recommend users who are more cautious and want greater rigidity to specify a small weight amount for their new versions (either roll it out gradually or immediately) and monitor traffic accordingly. If pleased, increase that weight a little bit and continue until full promote is desired.

PR #831, opened 2019-11-18
[0.6.x] CLI and controller updates for weights and rollouts

cbron:
https://github.com/rancher/rio/issues/774
https://github.com/rancher/rio/issues/719

This PR does several things: 

* breaks out any tricky math into functions and unit tests them, including fixes and tests for edge cases
* fixes bug where run command wouldn't de-promote other services
* add magic so rollout is smoother
* fixed general bugs around rollout controller and weights
* updates rollout duration to 4 seconds from 2, to help avoid conflicts in api
* add note to cli and docs that rollout duration is estimation

### DIFF
--- a/cli/cmd/create/create.go
+++ b/cli/cmd/create/create.go
@@ -69,7 +69,7 @@ type Create struct {
 	P_Ports                []string          `desc:"Publish a container's port(s) (format: svcport:containerport/protocol)"`
 	Privileged             bool              `desc:"Run container with privilege"`
 	ReadOnly               bool              `desc:"Mount the container's root filesystem as read only"`
-	RolloutDuration        string            `desc:"How long the rollout should take" default:"0s"`
+	RolloutDuration        string            `desc:"How long the rollout should take.  An approximation, actual time may fluctuate" default:"0s"`
 	RequestTimeoutSeconds  int               `desc:"Set request timeout in seconds"`
 	Scale                  string            `desc:"The number of replicas to run or a range for autoscaling (example 1-10)"`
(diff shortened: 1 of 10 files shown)


scripts/test
```diff
@@ -43,5 +43,6 @@ if [[ ${ARCH} == amd64 ]]; then
 
     trap print_log exit
 
-    go test -v ./tests/integration/... --integration-tests
+    echo 'Running integration tests...'
+    go test ./tests/integration/... --integration-tests
```
cbron: I changed behavior of test output so it wasn't so cluttered

cli/cmd/weight/weight.go
```diff
@@ -18,10 +19,13 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-const DefaultInterval = 2
+const (
+	DefaultInterval = 4
```
cbron: Default interval is now 4

cli/cmd/create/create.go
```diff
@@ -98,7 +98,22 @@ func (c *Create) RunCallback(ctx *clicontext.CLIContext, cb func(service *riov1.
 	}
 
 	service = cb(service)
-	return service, ctx.Create(service)
+	err = ctx.Create(service)
+	if err != nil {
+		return nil, err
+	}
+	if c.Weight == 100 {
+		err = weight.PromoteService(ctx, types.Resource{
```
StrongMonkey: Nit: 
```suggestion
		return service,  weight.PromoteService(ctx, types.Resource{
```

PR #834, opened 2019-11-19
Don't assume revision has fixed length

StrongMonkey:


StrongMonkey:
Flaky tests, merging.

Issue #780, opened 2019-11-14
Align p1 test plan and automation


rancher-max:
This is also a good opportunity to add to p1s as needed and update automated integration and validation tests. For the purposes at this point, it might just require adding TODOs unless it can be done quickly.

PR #835, opened 2019-11-19
[0.6.x] Move tests to more aptly-named files and fix flakes

rancher-max:
Closes: #780


tests/integration/weight_test.go
```diff
@@ -41,4 +35,30 @@ func weightTests(t *testing.T, when spec.G, it spec.S) {
...
+			stagedService = service.Stage("ibuildthecloud/demo:v3", "v3")
+			stagedService.Promote()
+		})
+
+		it("should retain all revision endpoints with an app endpoint pointing to the new version", func() {
+			assert.Equal(t, "Hello World", service.GetEndpointResponse())
+			assert.Equal(t, "Hello World v3", stagedService.GetEndpointResponse())
+			assert.Equal(t, 0, service.GetCurrentWeight())
+			assert.Equal(t, 100, stagedService.GetCurrentWeight())
+			//assert.Equal(t, 0, service.GetKubeCurrentWeight())
```
cbron: Should just delete these I think. We would need to do a full calc with kubectl to do it and that logic might change in rio which makes anything in tests brittle, so I don't think they are valuable. You can do in follow up PR.

tests/validation/weight_test.go
```diff
@@ -15,27 +15,25 @@ func weightTests(t *testing.T, when spec.G, it spec.S) {
...
+			// Sleeping 9 seconds to guarantee 5 rollout ticks with 1 second to spare since the default tick interval is 2 seconds.
+			stagedService.WeightWithoutWaiting(60, "--duration=1m")
 			time.Sleep(8 * time.Second)
-
-			stagedWeightAfter10Seconds := stagedService.GetCurrentWeight()
-			serviceWeightAfter10Seconds := service.GetCurrentWeight()
-			assert.Equal(t, 30, stagedWeightAfter10Seconds)
-			assert.Equal(t, 100, serviceWeightAfter10Seconds)
+			stagedComputedWeightAfter10Seconds, serviceComputedWeightAfter10Seconds := stagedService.GetComputedWeight(), service.GetComputedWeight()
+			assert.Equal(t, 20, stagedComputedWeightAfter10Seconds)
```
cbron: Does this actually work ? I'm worried that as I change the rollout config this will break, if we add a magic +- 50% scaler thing the ticks won't be equal anymore. Guess we can change later, but I think it would be better to do a > 10 < 30 or something. We can figure out later when I break it :)

rancher-max: I considered that. This did seem to work after I set time to 9, but when time was 8 I noticed weight would be at 17 about 30% of the time, and time at 10 the weight was 25 about 30% of the time. So this could've definitely been flaky. 
Fixed to your suggestion, and will update as needed based on your rollout config changes.

tests/validation/exec_test.go
```diff
@@ -0,0 +1,33 @@
...
+			otherService.Create(t, "ibuildthecloud/demo:v3")
+		})
+
+		it.After(func() {
+			service.Remove()
+			otherService.Remove()
+		})
+
+		it("should allow accessing services directly from within other service pods in the same namespace", func() {
+			//rio exec -it d-test@v1 /bin/sh -c "apt-get update && apt-get install -y --fix-missing curl && curl --version"
```
cbron: ?

rancher-max: Oh I can remove that comment. That's just the command it runs

PR #843, opened 2019-11-20
Generate new API docs, add custom domain

StrongMonkey:


docs/continuous-deployment.md
```diff
@@ -42,38 +42,51 @@ If the `--template` flag is not set, then every subsequent build will overwrite
...
 $ rio secret create --docker
 Select namespace[default]: $(put the same namespace with your workload)
 Registry URL[https://index.docker.io/v1/]: $(found with "docker info | grep Registry")
 username[]: $(your docker username)
 password[******]: $(your docker password)
 ```
-2. Create a workload pointing to your image. For example:
-`rio run --image-pull-secrets dockerconfig -p 8080 imageorg/imagename:version`
+
+2. Push your image to docker registry:
```
rancher-max: Can we also clarify that this also runs the built image?

StrongMonkey: @rancher-max Feel like runs the built image shouldn't be here. It will be in the section of `running workload` with just imagePullSecrets

rancher-max: @StrongMonkey I'm fine with that. In that case should we remove the similar "create a workload" parts elsewhere in this file?

StrongMonkey: Where is that? can you put suggestions in here? I saw other examples are just purely running http repo.

docs/continuous-deployment.md
```diff
@@ -42,38 +42,51 @@ If the `--template` flag is not set, then every subsequent build will overwrite
...
     1. Configure git sshkey auth credential secrets. This should use a key that does not have a password associated to it:
+       
         ```bash
         $ rio secret create --git-sshkey-auth
         Select namespace[default]: $(put the same namespace with your workload)
         git url[]: $(put your github url. Leave out http/https. Example: github.com)
         ssh_key_path[]: $(type the path to your ssh private key)
         ```
+       
     2. Create workload pointing to your repo using ssh checkout. For example:
```
rancher-max:
```suggestion
    2. Tell rio to do ssh checkout from git. For example:
```

docs/continuous-deployment.md
rancher-max: Okay I understand now. Made a few suggestions that might help other section clarity as well.

PR #852, opened 2019-11-21
Update example to include default Namespace value

rancher-max:


tests/integration/route_test.go
```diff
@@ -45,7 +45,6 @@ func routeTests(t *testing.T, when spec.G, it spec.S) {
 			assert.Equal(t, "Hello World v3", routeB.GetEndpointResponse())
 			assert.Equal(t, "Hello World", routeA.GetKubeEndpointResponse())
 			assert.Equal(t, "Hello World v3", routeB.GetKubeEndpointResponse())
-			assert.Equal(t, routeA.Router.Status.Endpoints, routeB.Router.Status.Endpoints)
```
rancher-max: @StrongMonkey What do you think about deleting this? The test failed because sometimes https takes a little longer to become available. Or should we add some sort of wait until https is ready before running integration tests? (cert manager pod?)

StrongMonkey: we can just compare `http` endpoints. Add a util method that extract http endpoint and compare

Issue #824, opened 2019-11-18
router ux updates


cbron:
* The router target is ambiguous when adding. Maybe add `(app)` or similar to the view to help distinguish.
*  target version should use `@` not colon

rancher-max:
Looks much nicer! Fixed in master

PR #853, opened 2019-11-21
Make router app vs version ux clearer

cbron:
Problem: if you do `rio ps` you get just `z` for `z@v0`, but if you do `rio route` then `z` means entire app not `z@v0`. What does everyone think of this solution ? 

router

https://github.com/rancher/rio/issues/824


cbron:
```0
0. z,port=80
1. z@app,port=80
2. z(app),port=80
3. z-app,port=80
4. z"app",port=80
5. z/app,port=80
```

I like 1 and 2 the best, 0 and 3 are misleading.

cbron:
went with 2

PR #857, opened 2019-11-22
Add integration test for local image build and run

rancher-max:


tests/integration/build_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/rancher/rio/tests/testutil"
+)
+
+func buildTests(t *testing.T, when spec.G, it spec.S) {
+
+	var service testutil.TestService
+
+	when("an image is built locally in rio", func() {
+		it.Before(func() {
+			service.CreateLocal(t)
```
cbron: `CreateLocal` is somewhat ambiguous when reading it here. How about `BuildAndCreate` or `CreateFromDockerfile`

tests/testutil/service.go
```diff
@@ -135,6 +135,43 @@ func (ts *TestService) Attach() []string {
...
+	return ts.Stage(fmt.Sprintf("localhost:5442/%s/%s:%s", TestingNamespace, imageName, imageVersion), imageVersion)
+}
+
+// BuildAndCreate builds a local image and runs a service using it
+func (ts *TestService) BuildImage(imageName string, imageVersion string, args ...string) error {
+	pwd, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("Could not retrieve working directory.  %v", err.Error())
+	}
+	if strings.Contains(pwd, "tests/integration") {
```
StrongMonkey: hmmm does pwd ever change?

rancher-max: Not in Drone 😬 

I found that for some reason, `pwd` was always `tests/integration` in Drone or running from the Dockerfile. The other case accounts for running locally which we tend to do from rio root directory.

PR #860, opened 2019-11-22
Make availablereplicas return if it is available to

rancher-max:


tests/testutil/service.go
```diff
@@ -284,7 +284,7 @@ func (ts *TestService) GetAppEndpointResponse() string {
 
 // Returns count of ready and available pods
 func (ts *TestService) GetAvailableReplicas() int {
-	if ts.Service.Status.DeploymentReady && ts.Service.Status.ScaleStatus != nil {
+	if ts.Service.Status.DeploymentReady && ts.Service.Status.ScaleStatus != nil || (&ts.Service.Status.DeploymentReady == nil && ts.Service.Status.ScaleStatus != nil) {
```
cbron: Why are you checking the address of `DeploymentReady` ? Can't you just check if true or false.

rancher-max: essentially it has 3 states in the manifest: `true`, `false`, or nonexistent (nil address). I want to get the scale status if it is true or nonexistent, because during scale up/down there can be available replicas and the service will be ready, but this field will not be present.

Issue #772, opened 2019-11-13
Tailing service log output should show container state in running pods


rancher-max:
**Describe the bug**
When tailing the logs for a service, it should continue to show the container log output for the destroyed pod and the newly created pod for all newly created/destroyed pods, not just the first two.
Refer to: https://github.com/rancher/rio/issues/614#issuecomment-553619362

Type:
**Rio version**: `v0.6.0-rc1`

cbron:
@rancher-max when you test this can you do it with a service that logs stuff to ensure it is flowing correctly between deletes. The trick to recreating bug is to delete pod so do that a few times while logging non-previous svcs and also (in another terminal) doing previous svcs. Make sure both work.

rancher-max:
```
$ rio logs focused-goldwasser
+ default focused-goldwasser-v0t65qb-799ccbd654-kjwz4 › focused-goldwasser-v0t65qb
focused-goldwasser-v0t65qb-799ccbd654-kjwz4 focused-goldwasser-v0t65qb 2019/11/26 20:26:58 Basic logging -- request received
focused-goldwasser-v0t65qb-799ccbd654-kjwz4 focused-goldwasser-v0t65qb 2019/11/26 20:27:01 Basic logging -- request received
focused-goldwasser-v0t65qb-799ccbd654-kjwz4 focused-goldwasser-v0t65qb 2019/11/26 20:27:02 Basic logging -- request received
focused-goldwasser-v0t65qb-799ccbd654-kjwz4 focused-goldwasser-v0t65qb 2019/11/26 20:27:13 Basic logging -- request received
+ default focused-goldwasser-v0t65qb-799ccbd654-gpr67 › focused-goldwasser-v0t65qb
Error opening stream to default/focused-goldwasser-v0t65qb-799ccbd654-gpr67: focused-goldwasser-v0t65qb
: container "focused-goldwasser-v0t65qb" in pod "focused-goldwasser-v0t65qb-799ccbd654-gpr67" is waiting to start: PodInitializing
- default focused-goldwasser-v0t65qb-799ccbd654-kjwz4

^C
$ rio logs focused-goldwasser
+ default focused-goldwasser-v0t65qb-799ccbd654-gpr67 › focused-goldwasser-v0t65qb
focused-goldwasser-v0t65qb-799ccbd654-gpr67 focused-goldwasser-v0t65qb 2019/11/26 20:27:42 Basic logging -- request received
[...]

Issue #708, opened 2019-11-07
Rio should throw a warning/error when doing rio logs on a service that has scale zero [v0.6.0-alpha.3]


izaac:
**Describe the bug**
When a service has SCALE zero, if we try to see the logs using `rio logs <service>` rio just sits there doing nothing. 

`rio exec <service> <command>` throws an error.

```bash
rio exec ascaler uname
FATA[0000] failed to find any pod for service ascaler, container ""
```

**To Reproduce**
1. create a service with auto scaling from 0 to X.
2. `rio ps` the service and check the scale is 0
3. `rio logs <service>`

rio logs just sits there without doing anything, I had to CTRL-C.

**Expected behavior**

Throw an error when using `rio logs` and scale is zero like with `rio exec` 

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.6", GitCommit:"96fac5cd13a5dc064f7d9f4f23030a6aeface6cc", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:49Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.6-eks-5047ed", GitCommit:"5047edce664593832e9b889e447ac75ab104f527", GitTreeState:"clean", BuildDate:"2019-08-21T22:32:40Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.6.0-alpha.3 (98e09134)
Rio CLI Version: v0.6.0-alpha.3 (98e09134)
Cluster Domain: xxxxx.on-rio.io
Cluster Domain IPs:
System Namespace: rio-system
Wildcard certificates: true
```
**Additional context**
`rio systemlogs`
output:
```
Nothing unusual on systemlogs.
```

rancher-max:
This now shows `Waiting for pods...` and then shows logs. However, it has the same issue as https://github.com/rancher/rio/issues/772#issuecomment-558804763 where it does not have log results:
```
Waiting for pods...
                                               
+ default gh-v0wrtkd-659f4c9fc4-fj566 › gh-v0wrtkd
Error opening stream to default/gh-v0wrtkd-659f4c9fc4-fj566: gh-v0wrtkd
[...]

PR #863, opened 2019-11-22
Error on zero-scaled service and allow logging after pod deletion

cbron:
https://github.com/rancher/rio/issues/708
https://github.com/rancher/rio/issues/772

### DIFF
--- a/cli/cmd/logs/logs.go
+++ b/cli/cmd/logs/logs.go
@@ -65,7 +65,7 @@ func (l *Logs) setupConfig(ctx *clicontext.CLIContext) (*stern.Config, error) {
 			return nil, errors.New("No object found")
 		}
 		if obj.Object.(*riov1.Service).Status.DeploymentReady == true && obj.Object.(*riov1.Service).Status.ScaleStatus != nil && obj.Object.(*riov1.Service).Status.ScaleStatus.Available == 0 {
-			return nil, errors.New("Service has scale of 0")
+			fmt.Println("Waiting for pods...")
 		}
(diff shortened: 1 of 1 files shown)


cli/cmd/logs/logs.go
```diff
@@ -63,6 +64,9 @@ func (l *Logs) setupConfig(ctx *clicontext.CLIContext) (*stern.Config, error) {
 		if obj.Object == nil {
 			return nil, errors.New("No object found")
 		}
+		if obj.Object.(*riov1.Service).Status.DeploymentReady == true && obj.Object.(*riov1.Service).Status.ScaleStatus != nil && obj.Object.(*riov1.Service).Status.ScaleStatus.Available == 0 {
```
StrongMonkey: Why do we need to throw an error? Is there a use case that we can wait for pod to be created? Feel like it is not really needed. If service has zero pod just log nothing cause there is no pod.

cbron: 708 was made to bring inline with `exec`, but I agree it could go either way. It's slightly confusing because there is no output, it looks like command is hanging. Perhaps we could log "no current pods to log" or something.

StrongMonkey: I just think there is a use case to log service from the very beginning and wait for pods to be created. If this change is made then you can log service only if there are pods that already exist.

cbron: Yes, lets print and wait then: `Waiting for pods....` and then continue logging.

cbron: updated

PR #865, opened 2019-11-24
use same logic to read content as defaultfile, even if its a custom f…

### LINKED ISSUE #864: go templated custom rio file fails to bring up stack.
**Describe the bug**

**To Reproduce**
1. 

Try the example riofile . This will work if the name of the file is the default, which is Riofile

It fails to bring up anything if the name is different

```
services:
{{- if eq .Values.FOO "BAR" and eq .Values.NAMESPACE "test" }}
  demo:
    image: ibuildthecloud/demo:v1
    ports:
    - 80
{{- end}}

template:
  goTemplate: true # use go templating
  envSubst: true # use ENV vars during templating  
  questions:  # now make some questions that we provide answers too
  - variable: FOO
    description: ""


```

**Expected behavior**

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```

```
Type:
**Rio version**: `rio info`
```

```
**Additional context**
`rio systemlogs`
output:
```

```

lucidprogrammer:
fixes #864

### DIFF
--- a/cli/pkg/up/build.go
+++ b/cli/pkg/up/build.go
@@ -14,7 +14,6 @@ import (
 	riov1 "github.com/rancher/rio/pkg/apis/rio.cattle.io/v1"
 	"github.com/rancher/rio/pkg/constants"
 	"github.com/rancher/rio/pkg/stack"
-	"sigs.k8s.io/yaml"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cbron:
Hey @lucidprogrammer, thanks for the issue. We are going to fix with a different approach (PR linked in issue) as we want to keep the functionality of the local `readFile` method, and we found other surrounding issues.

PR #867, opened 2019-11-25
Named riofiles shouldnt fallback to default riofile

cbron:
Issue: https://github.com/rancher/rio/issues/864

### Fixes
1. Doesn't fallback to `Riofile` when a named riofile fails
2. Doesn't unmarshall and error a named Riofile, as non-named Riofiles don't
3. Fixes panic
4. Fixes Readme

### Testing

Number 1. 
Have 2 Riofiles, one named Riofile and one named something else with different data that will fail to parse. Run `rio up -f named_file.yaml`. 

Number 2. 
Setup: save this as anything but `Riofile` and run up with `-f`, should fail to load, but it does load if named `Riofile`. Fix should allow either way. 
```
{{- if (and (eq .Values.NAMESPACE "test") (eq .Values.FOO "BAR")) }}
services:
  demo:
    image: ibuildthecloud/demo:v1
    ports:
    - 80
{{- end}}

template:
  goTemplate: true # use go templating
  envSubst: true # use ENV vars during templating  
  questions:  # now make some questions that we provide answers too
  - variable: FOO
    description: "My custom thing"
  - variable: NAMESPACE
    description: "The namespace"
```

Number 3. 

Name this Riofile and run just `rio up`
```
services:
{{- if eq .Values.FOO "BAR" and eq .Values.NAMESPACE "test" }}
  demo:
    image: ibuildthecloud/demo:v1
    ports:
    - 80
{{- end}}
```

### DIFF
--- a/cli/pkg/up/build.go
+++ b/cli/pkg/up/build.go
@@ -1,6 +1,7 @@
 package up
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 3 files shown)


cbron:
@StrongMonkey check part 2 especially, the removed error check. I'm not sure why that was in there but a standard Riofile doesn't do it.

cli/pkg/up/build.go
```diff
@@ -62,10 +61,7 @@ func LoadRiofile(path string) ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
-		var r map[string]interface{}
-		if err := yaml.Unmarshal(content, &r); err == nil {
-			return content, nil
-		}
+		return content, nil
```
StrongMonkey: We also need this workflow to work:
If `rio up -f Dockerfile`, use the defaultRiofileContent. This is just for convenience if user pass a dockerfile.(darren want this feature).

StrongMonkey: So we might have to apply answer first and check if it is a yaml manifest, otherwise we treat it as Dockerfile and will fall back to look for Riofile.

pkg/template/parse.go
```diff
@@ -104,6 +105,9 @@ func (t *Template) parseContent(answersCB AnswerCallback) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	if template == nil {
```
StrongMonkey: why need this error? We have to be carefully not throwing error when template is invalid. We should just fall back and continue.

StrongMonkey: This will break if user defines the following Riofile:
```yaml
services:
  demo:
    image: foo
    template: true
```

cli/pkg/up/build.go
cbron: Check `LoadRiofile` now and see if you think that will work for all cases

pkg/template/parse.go
cbron: That actually works for me with changes, I tested a few different files to see if I could cause error on good Riofile and couldn't. But I'll just have it return content to be safe.

cli/pkg/up/build.go
cbron: If you think that works we should prob add a unit test

PR #868, opened 2019-11-25
Update gloo to 1.0.0

StrongMonkey:


StrongMonkey:
Need to wait for https://github.com/solo-io/gloo/issues/1716 to be solved. Currently running on a custom version of gloo.

StrongMonkey:
Also need to add migration path for upgrading.

daxmc99:
Is this a non-breaking upgrade?

StrongMonkey:
This is actually a breaking change upgrade in gloo, but we takes care of gloo upgrade so there shouldn't need any manual operations. Doesn't guarantee zero downtime.

modules/gloo/features/feature.go
```diff
@@ -38,7 +55,57 @@ func Register(ctx context.Context, rContext *types.Context) error {
...
+			// handle upgrade
+			return upgradeOldVirtualService(rContext.Namespace, rContext.RestConfig)
+		},
 	}
 
 	return feature.Register()
 }
+
+// upgradeOldVirtualService handles compatibility from upgrading gloo from v0.20.x to v1.0.0 for virtualservices.
+func upgradeOldVirtualService(systemNamespace string, config *rest.Config) error {
```
cbron: This says upgrade, but really it just deletes them so they can be recreated ?

StrongMonkey: Corrrect, delete non-compatible types

Issue #684, opened 2019-11-04
`rio route` and `rio services` doesn't support timeout 


StrongMonkey:
Right now there is no way to configure timeout setting to gloo gateway for routes and services.

StrongMonkey:
This is not urgent to test for v0.6.0. We basically added parameter for router and services you can configure timeout for each rule.

rancher-max:
`rio run --request-timeout-seconds 1 -n ghout -p 8080 https://github.com/rancher-max/rio-demo' causes a panic in the rio-controller and it becomes broken. Had to uninstall rio and reinstall to get it working again.

```
rio-controller | E1107 18:45:33.311015       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
rio-controller | goroutine 873 [running]:
rio-controller | github.com/rancher/rio/vendor/k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1c13a80, 0x3433e10)
rio-controller | 	/go/src/github.com/rancher/rio/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0xa3
rio-controller | github.com/rancher/rio/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
rio-controller | 	/go/src/github.com/rancher/rio/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x82
rio-controller | panic(0x1c13a80, 0x3433e10)
rio-controller | 	/usr/local/go/src/runtime/panic.go:679 +0x1b2
rio-controller | github.com/rancher/rio/modules/gloo/pkg/vsfactory.(*VirtualServiceFactory).ForRevision(0xc000ebcc00, 0xc0005ac000, 0xc000c9e180, 0x1fd7c90, 0xc00224ab98, 0xc00224aad0, 0xc00224ac78)
rio-controller | 	/go/src/github.com/rancher/rio/modules/gloo/pkg/vsfactory/revision.go:23 +0x5c1
rio-controller | github.com/rancher/rio/modules/gloo/controller/service.(*handler).generate(0xc0006302a0, 0xc0005ac000, 0x0, 0x0, 0xc0008f648b, 0x5, 0xc0008f6590, 0x2, 0x0, 0xc001773250, ...)
[...]

PR #869, opened 2019-11-26
Fix panic when setting request timeout

StrongMonkey:


StrongMonkey:
Fixed https://github.com/rancher/rio/issues/684.

PR #871, opened 2019-11-26
Change dashboard name

StrongMonkey:


cli/cmd/dashboard/dashboard.go
```diff
@@ -16,6 +16,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	dashboardAdmin = "dashboard-admin"
```
cbron: Should we call this `rio-dashboard-admin` just to be explicit ? I doubt we would get a conflict with this name but it might not be immediately obvious what its for.

Issue #828, opened 2019-11-18
Reduce Rio cli size


davidnuzik:
The Rio CLI is roughly the size of K3s.
Is there anything we can do to reduce the file size? Needs investigation and work.

PR #882, opened 2019-12-02
Move package around, reduce binary size

StrongMonkey:
This commit moves package files to a common place so that CLI code will
not import package code inside controller


cbron:
Issue: https://github.com/rancher/rio/issues/828

PR #898, opened 2019-12-06
Make installation lighter by default, add ingress mode

StrongMonkey:
This commit support ARM, and turn on autoscaling, build on demand, and
disable service mesh if total memory is less than 3GB.


StrongMonkey:
Will need to update cert-manager image.

StrongMonkey:
Actually This PR will not include ARM(we have to deal with cert-manager image and also our own implementation of ACME). So this PR will focus to providing on demand feature start on autoscaling and build, add ingress mode back, and some minor tweaks.

pkg/server/startup.go
```diff
@@ -56,6 +65,46 @@ func bootstrapResources(rioContext *types.Context, systemNamespace string) error
...
+		f.Enabled = &[]bool{true}[0]
+		conf.Features["*"] = f
+	} else {
+		for _, feature := range strings.Split(config.ConfigController.Features, ",") {
+			if feature == "" {
+				continue
+			}
+			f := conf.Features[feature]
+			f.Enabled = &[]bool{true}[0]
+			conf.Features[feature] = f
```
cbron: Why can't you just do `conf.Features[feature].Enabled =...` ?

main.go
```diff
@@ -67,6 +67,12 @@ func main() {
 			Usage:       "Specify which ip address RDNS should generate record for",
 			Destination: &config.ConfigController.IPAddresses,
 		},
+		cli.StringFlag{
+			Name:        "features",
+			Usage:       "Specify which feature must be turned on",
```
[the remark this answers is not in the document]
StrongMonkey: This is to add a backdoor in controller binary arguments to fix tests. Since we disable build and autoscaling by default, It will break CI. So by default if you are running in binary mode we enable build and autoscaling.

modules/linkerd/feature/feature.go
```diff
@@ -49,5 +60,27 @@ func Register(ctx context.Context, rContext *types.Context) error {
 			return nil
 		},
 	}
+
+	// only upgrade linkerd when configmap upgrade option is true
```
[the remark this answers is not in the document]
StrongMonkey: sure I will make an issue and link that.

pkg/server/startup.go
StrongMonkey: I don't think golang allows you to do that.

StrongMonkey:
Also address https://github.com/rancher/rio/issues/908.

modules/service/controllers/service/controller.go
```diff
@@ -47,17 +53,66 @@ type serviceHandler struct {
...
+		conf.Features["autoscaling"] = f
+	}
+
+	for _, con := range services.ToNamedContainers(service) {
+		if con.ImageBuild != nil && con.ImageBuild.Repo != "" && arch.IsAmd64() {
+			if conf.Features == nil {
+				conf.Features = map[string]config.FeatureConfig{}
+			}
+			f := conf.Features["build"]
+			f.Enabled = &[]bool{true}[0]
```
daxmc99: I don't love this but... I think its the only way to do this in a single line....maybe https://stackoverflow.com/questions/28817992/how-to-set-bool-pointer-to-true-in-struct-literal

modules/service/controllers/service/controller.go
```diff
@@ -47,17 +53,66 @@ type serviceHandler struct {
...
+	if err != nil {
+		return err
+	}
+
+	if services.AutoscaleEnable(service) && arch.IsAmd64() {
+		if conf.Features == nil {
+			conf.Features = map[string]config.FeatureConfig{}
+		}
+		f := conf.Features["autoscaling"]
+		f.Enabled = &[]bool{true}[0]
```
daxmc99: same as below

cli/cmd/install/install.go
```diff
@@ -162,23 +163,12 @@ func (i *Install) preConfigure(ctx *clicontext.CLIContext, ignoreCluster bool) e
...
 	versionWarning := false
 	for _, node := range nodes.Items {
-		totalMemory += node.Status.Capacity.Memory().Value()
 		if !versionWarning {
 			versionWarning = checkKubernetesVersion(node.Status.NodeInfo.KubeletVersion)
 		}
 	}
-	if totalMemory < 2147000000 {
-		if isMinikubeCluster(nodes) {
-			fmt.Println("Warning: detecting that your minikube cluster doesn't have at least 3 GB of memory. Please try to increase memory by running `minikube start --memory 4098`")
```
daxmc99: Our install is pretty quiet right now, maybe we should mention we are installing components as they are needed.

StrongMonkey: yep, we should definitely call this out in docs and release notes.

pkg/server/startup.go
```diff
@@ -56,6 +65,46 @@ func bootstrapResources(rioContext *types.Context, systemNamespace string) error
...
+		f := conf.Features["*"]
+		f.Enabled = &[]bool{true}[0]
+		conf.Features["*"] = f
+	} else {
+		for _, feature := range strings.Split(config.ConfigController.Features, ",") {
+			if feature == "" {
+				continue
+			}
+			f := conf.Features[feature]
+			f.Enabled = &[]bool{true}[0]
```
daxmc99: Nit here as well

pkg/server/startup.go
```diff
@@ -56,6 +65,46 @@ func bootstrapResources(rioContext *types.Context, systemNamespace string) error
...
+	if err != nil {
+		return err
+	}
+
+	if conf.Features == nil {
+		conf.Features = map[string]config.FeatureConfig{}
+	}
+	if config.ConfigController.Features == "*" {
+		f := conf.Features["*"]
+		f.Enabled = &[]bool{true}[0]
```
daxmc99: Nit here as well

Issue #906, opened 2019-12-09
Registering publicdomain to app does not work if there is no v0 version


rancher-max:
**Describe the bug**
Running `rio domain register foo.bar myapp` fails if I don't have a `v0` of `myapp`

**To Reproduce**
1. `rio run -n myapp@v1 -p 80 nginx`
2. `rio domain register foo.bar myapp`

**Expected behavior**
This should register the app to foo.bar. This only works if there exists a `v0` on the app.

rancher-max:
Fixed!

PR #910, opened 2019-12-09
Fix target can't be found if target is app

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rio/issues/906.

cli/cmd/publicdomain/add.go
```diff
@@ -22,30 +22,41 @@ func (r *Register) Run(ctx *clicontext.CLIContext) error {
...
-		svc := result.Object.(*riov1.Service)
-		targetApp, targetVersion := services.AppAndVersion(svc)
-		if !strings.Contains(target, "@") {
-			targetVersion = ""
+
+	/*
+		1) if target is [namespace:]name and can't be found as an object, treat it as app
+		2) if target is [namespace:]name@version, treat it as service(revision)
+		3) if target is [namespace:]name and can be found as an object, treat it as a router
+	*/
```
cbron: Nice comments. Would be nice to add unit or integration tests to this in long term.

StrongMonkey: We can cover that in validation tests. I think all public domain tests are validation tests.

cbron: cc @rancher-max ^ low priority, similar to rio build cases.

Issue #849, opened 2019-11-21
Rio Info should show all available cluster domains


rancher-max:
After adding a cluster domain to the rio-system namespace, I should be able to see it when running `rio info` in a comma separated list, similar to Cluster IPs. 

Also, wildcard certificates should show as `true` when using a wildcard domain. 

Reference: rancher/rio#700

rancher-max:
Looks good now. `rio info` shows comma separated Cluster Domains and Wildcard Certificates with either `(true)` or `(false)` at the end of each of them if they are wildcard domains
```
$ rio info
Rio Version: 24c33d61-dirty (24c33d61)
Rio CLI Version: dev (HEAD)
Cluster Domain: gagb1s.on-rio.io,rio.qa.rancher.space
Cluster Domain IPs: 167.172.121.222
System Namespace: rio-system
Wildcard certificates: gagb1s.on-rio.io(true),rio.qa.rancher.space(true)
```

Issue #850, opened 2019-11-21
Endpoints for custom cluster domains only show http even if https is available


rancher-max:
Both the http and https endpoints should be available when using custom cluster domains and specifying both. It should default to showing https in the cli.

Current state does not show the https endpoint at all, and it's not in the manifest when using `rio inspect` to get endpoints.

Reference: rancher/rio#700

rancher-max:
https is shown when it is specified and a secret is supplied. When a secret is not supplied, https is not supported and then only http is shown.

mbarlow:
@rancher-max can you give more details for: https is shown when it is specified and a secret is supplied? How is https specified and how is secret supplied? I followed the instructions here:  https://github.com/rancher/rio/blob/master/docs/faq.md at bottom creating a clusterdomain with the entry secretName set to my wildcard tls secret and `rio info` still shows (false) for my wildcard and deploying a workload w/ rio still only shows http for endpoint when doing `rio ps` or `rio inspect`. I am using v0.7.0.

rancher-max:
@mbarlow We did some investigating on this. Following the steps from here: https://github.com/rancher/rio/issues/700#issuecomment-572414602 if you just do a bare `rio install` OR `rio install --disable-features rdns` it should work. I'm noticing that for some reason it does not work when we disable letsencrypt as it appears to only set ClusterDomain to https inside the LE controller. 

Will use https://github.com/rancher/rio/issues/990 to enhance this functionality.

PR #911, opened 2019-12-09
Add custom domain and certs in CLI

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rio/issues/849 and https://github.com/rancher/rio/issues/850.

cli/cmd/info/info.go
```diff
@@ -45,10 +52,10 @@ func info(ctx *clicontext.CLIContext) error {
 
 	builder.WriteString(fmt.Sprintf("Rio Version: %s (%s)\n", info.Status.Version, info.Status.GitCommit))
 	builder.WriteString(fmt.Sprintf("Rio CLI Version: %s (%s)\n", version.Version, version.GitCommit))
-	builder.WriteString(fmt.Sprintf("Cluster Domain: %s\n", clusterDomain.Name))
+	builder.WriteString(fmt.Sprintf("Cluster Domain: %v\n", strings.Join(names, ",")))
 	builder.WriteString(fmt.Sprintf("Cluster Domain IPs: %s\n", strings.Join(addresses, ",")))
 	builder.WriteString(fmt.Sprintf("System Namespace: %s\n", info.Status.SystemNamespace))
-	builder.WriteString(fmt.Sprintf("Wildcard certificates: %v\n", clusterDomain.Status.HTTPSSupported))
+	builder.WriteString(fmt.Sprintf("Wildcard certificates: %v\n", strings.Join(certs, ",")))
```
cbron: This will be a list of booleans ? Should we add additional context so they can see which are true/false ?

modules/letsencrypt/controllers/clusterdomain/controller.go
```diff
@@ -66,6 +66,7 @@ func (f *certsHandler) onCertChange(key string, obj *certmanagerv1alpha2.Certifi
 func (f *certsHandler) Handle(obj *v1.ClusterDomain, status v1.ClusterDomainStatus) ([]runtime.Object, v1.ClusterDomainStatus, error) {
 	if obj.Spec.SecretName != "" {
 		status.AssignedSecretName = obj.Spec.SecretName
+		status.HTTPSSupported = true
```
daxmc99: Why do we need to set this? Shouldn't `status.HTTPSSupported = condition.Cond("Ready").IsTrue(cert)` on line 85 cause this to switch automatically?

StrongMonkey: If user specify secretName we will not request letsencrypt certs so we can't use that cert condition to set `HTTPSSupported`. This means user is managing secret by their side.

cli/cmd/info/info.go
StrongMonkey: it kind of matches the order of domains. Not sure what is the best format to convey this.

cbron: can we loop domains to label: `certificates: domain-A true, domain-B false`

Issue #765, opened 2019-11-13
rio --system-namespace is stuck during installation [v0.6.0-rc1]


izaac:
**Describe the bug**
use the `--system-namespace` global option for rio install. rio takes a long time and it get stuck here:

```bash
rio --system-namespace riosystemtest install
Deploying Rio control plane....
| Waiting for system components. Not ready deployments: [rio-system/autoscaler rio-system/buildkitd rio-system/cert-manager ...]
```
**To Reproduce**
1. `rio --system-namespace riosystemtest install`
2. wait until the installation progress
3. Cancel with Ctrl-C the install seems to be finished but deploying services is not working.

**Expected behavior**

Get rio installed in a different namespace and functional.

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.8-gke.12", GitCommit:"188432a69210ca32cafded81b4dd1c063720cac0", GitTreeState:"clean", BuildDate:"2019-11-07T19:27:01Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.6.0-rc1 (1dec45ce)
Rio CLI Version: v0.6.0-rc1 (1dec45ce)
Cluster Domain: xxxxxx.on-rio.io
Cluster Domain IPs: x.x.x.x
System Namespace: riosystemtest
Wildcard certificates: true
```
**Additional context**
`rio systemlogs`
output:
```
rio-controller | time="2019-11-13T16:32:54Z" level=info msg="Starting rio-controller, version: v0.6.0-rc1, git commit: 1dec45ce"
rio-controller | time="2019-11-13T16:32:54Z" level=info msg="Updating CRD services.rio.cattle.io"
rio-controller | time="2019-11-13T16:32:54Z" level=info msg="Updating CRD stacks.rio.cattle.io"
rio-controller | I1113 16:32:54.463657       1 leaderelection.go:241] attempting to acquire leader lease  riosystemtest/rio...
[...]

PR #913, opened 2019-12-09
Don't hardcode rio-system namespace

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rio/issues/765.

Issue #918, opened 2019-12-10
Build image should support image tag being set


cbron:
This riofile does not work because it tries to add a 2nd image tag to the end of `:dev`
```
services:
  revs:
    version: v1
    ports:
    - 8080
    build:
      branch: master
      buildContext: .
      dockerfile: Dockerfile
      repo: https://github.com/cbron/mybusybox
      imageName: ns/name:dev
```

rancher-max:
Working in master

PR #921, opened 2019-12-10
allow user to specify custom build tag

cbron:
Issue: https://github.com/rancher/rio/issues/918

### DIFF
--- a/modules/build/controllers/service/service.go
+++ b/modules/build/controllers/service/service.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


modules/build/controllers/service/service.go
```diff
@@ -139,13 +140,11 @@ func (p populator) populateBuild(buildKey, namespace string, build *riov1.ImageB
...
 	}
-
-	rev := build.Revision
-	if rev == "" {
+	if build.Revision == "" {
 		return nil
 	}
 
-	trName := name.SafeConcatName(buildKey, name.Hex(build.Repo, 5), name.Hex(rev, 5))
+	trName := name.SafeConcatName(buildKey, name.Hex(build.Repo, 5), name.Hex(build.Revision, 5))
```
cbron: note the taskrun name is still using revision, as does the rev passed into task git resource

PR #928, opened 2019-12-11
Fix a few of the current flakey tests

rancher-max:


tests/testutil/service.go
```diff
@@ -216,6 +216,9 @@ func (ts *TestService) Weight(weightSpec int, args ...string) {
 		if err != nil {
 			ts.T.Fatal(err.Error())
 		}
+		if weightSpec == 100 {
+			time.Sleep(5 * time.Second)
+		}
```
cbron: Whats this for ? Also weight spec is often set to 10,000 now depending on case and state of service, the default changed to that from 100.

rancher-max: This just calls `rio weight {service}={weightSpec}` where weightSpec should be a percentage, so this was to mimic the same change in Promote.

The overall hope of the sleep is to get around the Gloo issue where weights don't work properly for the first few seconds. I want a better way than just sleeping but I don't see one at this time. Also 5 seconds won't always work but hopefully it helps reduce this error coming up in our tests significantly 🤞

cbron: Why only on 100 ?

rancher-max: Because on 100 any subsequent calls would expect ONLY one service to be available, whereas on anything less than 100, we would expect both services to be available so any weird gloo things shouldn't be an issue in those cases. 

Maybe it should be on 0 as well? But we don't have any tests that set weight=0 at this time.

Issue #542, opened 2019-10-01
Webhook to GH Tags


davidnuzik:
We need to add the ability to do rio image building based on github tags (like we do on commits and PRs).

rancher-max:
Webhook is working and regression passes. The only thing that doesn't work (which also does not work in the current released version) is that after you set a githubtoken secret, it uses that by default on all git repos, so you can no longer pull and build repos that you don't have admin access on.

cbron:
^ Fixing that with: https://github.com/rancher/rio/issues/962

PR #929, opened 2019-12-12
Tag webhook support

cbron:
Issue: https://github.com/rancher/rio/issues/542

First commit also adds shorthand `-n` to rio up.

Testing scenarios, I already tested all but need to go through again: 
* tags on or off, ie: tag-only vs commit-only webhooks
* using the tag-query and tag-exclude params
* tags on a non master branch still deploy - don't really need to test as long as tags work in general, this would just be testing github functionality
* Using CLI to try to make a service with tag and branch or tag without webhook should fail
* tags on a templated service should still work
* using riofile (see below)
* tags and filters on a riofile stack `rio up -n demo --build-tag=true --build-tag-include=^v --build-tag-exclude=alpha  https://github.com/....`

Testing instructions:
1. setup cluster
2. `rio secret add --github-webhook`
3. `rio run -p 8080 -n tag-demo --build-webhook-secret=githubtoken --build-tag=true --build-tag-include="^v" --build-tag-exclude="alpha" https://github.com/cbron/mybusybox`
4. push new tag or new commit, see that new pod is created for that service

For using local cluster, between steps 2 and 3 above do:
* expose your local machine
* `rio domain register external_endpoint rio-system:webhook`
* Do step 3
* update webhook to point to your external_endpoint

Example riofile: 
```
services:
  demo:
    build:
      repo: https://github.com/cbron/mybusybox
      branch: master
      webhookSecretName: githubtoken
      tag: true
      tagInclude: ^v
      tagExclude: rc
    ports:
    - 8080
```


cli/cmd/create/build.go
```diff
@@ -21,6 +21,9 @@ func (c *Create) setBuildOrImage(imageName string, spec *riov1.ServiceSpec) erro
 			PushRegistrySecretName: c.BuildDockerPushSecret,
```
cbron: I changed filename ^

cli/cmd/create/build.go
```diff
@@ -9,6 +10,16 @@ import (
 
 func (c *Create) setBuildOrImage(imageName string, spec *riov1.ServiceSpec) error {
 	if services.IsRepo(imageName) {
+
```
StrongMonkey: Extra space?

modules/build/controllers/stack/stack.go
```diff
@@ -310,7 +310,7 @@ func populateWebhookAndSecrets(webhookService *riov1.Service, stack *riov1.Stack
 			RepositoryURL:                  stack.Spec.Build.Repo,
 			Enabled:                        true,
 			Push:                           true,
-			Tag:                            true,
+			Tag:                            false,
```
StrongMonkey: @cbron I think we should also support stack on tag.

cbron:
@rancher-max ping me if you need help testing, and try to review docs as you go, I think I got everything but not positive.

Issue #889, opened 2019-12-04
Change cert@rancher.dev and require users to explicitly set an email on rio install


rancher-max:
The ask came from Let's Encrypt:

> The cert@rancher.dev default is particularly problematic because it is not a deliverable address, and as far as I can tell rancher.dev is not actually owned by Rancher. Could I ask you to remove the default email address from Rio and require that users explicitly set an email before Rio creates an account?

We propose adding an (optional) flag for a user to supply a valid email address.
We may want to update docs as well to explain the purpose of the email flag and that if you want expiry notifications.

vincent99:
FWIW there's no MX record but it is a domain we own.

jsha:
Hi,

I hope you can fix this issue soon. At Let's Encrypt we're likely to start blocking new account creation from this address soon since it is undeliverable.

vincent99:
@jsha We value the service you are providing and would like to keep you guys happy.  I don't know exactly what all your concerns are yet so I would like to here more about what the problem is from your perspective, but here are a few ideas that come to mind that we could do anyway.

1) I could add a MX record and make `cert@rancher.dev` "deliverable" today.  Since every user still shares the same email it's not clear that this accomplishes much for anybody (i.e. I'd just deliver them to /dev/null).  But if the bounces themselves are a problem on your side we can do this.

2) Is providing an email to you actually required?  https://letsencrypt.org/docs/expiration-emails/ says "**If** you provide an email".  If permitted, the next version of Rio could just not provide any email address to you at all.

3) Each Rio install gets a random 6-character ID which we create a DNS record for `XXXXXX.on-rio.io` and then request a `*.XXXXXX.on-rio.io` cert from you.  So we could submit the email to you as `XXXXXX@on-rio.io`, and make those all accept mail.  These would still be "delivered" to /dev/null, because Rio is open-source and runs on random user's hardware and we have no idea who they are.  But the net result would be that each installation/user gets its own email address instead of all sharing one, if that helps you.

4) We could add a flag or installation prompt for the user to provide their email.  This seems totally reasonable as an _option_, as I'm sure some users would actually want their expiration notices.  I would prefer to be able to fall back to _something_ deterministic (like one of the options above) rather than **requiring** that users provide an email though.  There's no practical way for us to validate them so if we require input instead of opting-in you will end up with a good fraction that are random garbage.  And apparently your side is also doing no validation, since you take our current value one that has no MX record...

davidnuzik:
[...]

PR #931, opened 2019-12-12
Add prompt letencrypt email in install

StrongMonkey:


cli/cmd/install/install.go
```diff
@@ -203,6 +207,36 @@ func (i *Install) getConfigMap(ctx *clicontext.CLIContext, ignoreCluster bool) (
...
+				return cfg, nil, err
+			}
+			if num == 0 {
+				email, err := questions.Prompt("Please enter your email address:\n", "your@company.com")
+				if err != nil {
+					return cfg, nil, err
+				}
+				cfg.LetsEncrypt.Email = email
+			} else if num == 1 {
+				logrus.Warnf("No email is provided. This is strongly discouraged and lack of notifications may cause you to lose access to your certificates. Run `rio systemconfig --edit` to add it back")
```
cbron: Are we sure adding an email well after setup has completed will actually send it to LE ?

StrongMonkey: what do you mean?

cbron: If I install rio with option 2, and then 4 weeks later I update the config to include my email address, will that email get sent to letsencrypt and attached to that existing cert ?

StrongMonkey: I don't think it will take effect immediately, letsencrypts will probably have that information next time it renews certs. But we have nothing to do on our side to change this.

StrongMonkey:
Address https://github.com/rancher/rio/issues/889.

Issue #922, opened 2019-12-10
A few issues with rio lite install


rancher-max:
Changes to rio install have caused a few issues and annoyances:

**Issues:**
- `rio --system-namespace <value> install` does not map the `-s` flag to the newly chosen system namespace. (https://github.com/rancher/rio/issues/765#issuecomment-564275314)
- Disabling gloo feature does not properly setup ingress (needs http/https port to be set in the clusterdomain)
- Disabling linkerd (which happens by default on cluster with mem < 3gb) causes autoscaling not to work, even though autoscaler is enabled (this likely will be just a documentation thing).

**Annoyances:**
- `rio image` and `rio build-history` give fatal errors instead of something descriptive. (`FATA[0000] buildkitd pod not found` and `FATA[0000] the server could not find the requested resource (get taskruns.tekton.dev)`. Would be better if it said something like `Build feature is not enabled`)
- Editing installation features is a bit of a pain. It is json, but all squished onto one line.

rancher-max:
Running `rio up` or `rio build` both do not turn "Build" feature on.
`rio build` gives error: `FATA[0000] resource name may not be empty`
`rio up` creates a stack but then doesn't do anything with the Build feature so nothing actually happens

StrongMonkey:
Available to test on master now.

rancher-max:
Should `rio uninstall` default to whatever the systemnamespace value is? Right now it defaults to `rio-system` so to uninstall after `rio --system-namespace mytest install` I have to do: `rio uninstall --namespace mytest`

rancher-max:
2 things:
1. `rio image` should enable build feature like `rio build-history` does.

2. Running `rio install --disable-features gloo` after doing a `rio uninstall` does not generate any clusterdomain. Output from the command after it finishes:
```
$ rio install --disable-features gloo
Deploying Rio control plane....
Warning: clusterDomain is not generated
rio controller version dev (HEAD) installed into namespace rio-system
Controller logs are available from `rio systemlogs`

Welcome to Rio!

Run `rio run -p 80:8080 https://github.com/rancher/rio-demo` as an example
```
Nothing useful in systemlogs. Nothing from `kubectl get clusterdomains`, and `rio info` returns an error: `FATA[0000] no system information is generated`

rancher-max:
Looks good in master now!

PR #933, opened 2019-12-13
Fix uninstall tied to one namespace, fix rio images, add all features in configmap

StrongMonkey:


pkg/controllers/config/controller.go
```diff
@@ -167,5 +170,41 @@ func (f *featureHandler) onChange(key string, obj *corev1.ConfigMap) (*corev1.Co
...
+		conf.Features = map[string]config.FeatureConfig{}
+	}
+
+	state := f.featureState
+
+	t := true
+	for _, feature := range features.GetFeatures() {
+		f := conf.Features[feature.Name()]
+		if state.Has(feature.Name()) {
+			f.Enabled = &t
```
cbron: Trying to think if there is a case where using same address for all the true's will flip all at once. Think its ok, since t never changes and gets re-instantiated every call. I like this a lot better than the slice method.

StrongMonkey:
Address https://github.com/rancher/rio/issues/922.

PR #940, opened 2019-12-16
Fix build, domain, and validation tests

rancher-max:
This also enables building rio from master. So running the following will run all integration and validation tests in a fresh cluster:
```bash
$ docker build -f ./tests/Dockerfile -t rio-test .
$ docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -e test=all -e RIO_VERSION=master -e REPO=${DOCKER_USERNAME} -e TAG=dev -e CLUSTER=k3s -e TOKEN=${DigitalOcean_Token} -e WORKERS=2 -e AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY} -e AWS_SECRET_ACCESS_KEY=${AWS_SECRET_KEY} -e RIO_ROUTE53_ZONEID=${ROUTE53_ZONEID} -e RIO_ROUTE53_ZONENAME=${ROUTE53_ZONENAME} rio-test:latest
```

### DIFF
--- a/tests/validation/autoscale_test.go
+++ b/tests/validation/autoscale_test.go
@@ -25,7 +25,7 @@ func autoscaleTests(t *testing.T, when spec.G, it spec.S) {
 		it("should autoscale down to 0", func() {
 			// Precondition
 			assert.True(t, service.IsReady())
-			assert.Equal(t, "Hello World", service.GetAppEndpointResponse())
+			assert.Equal(t, "Hi there, I am rio:production6", service.GetAppEndpointResponse())
 
 			// When no requests happen for a while, it should scale to 0
 			service.WaitForScaleDown()
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:1.13-alpine
 # Install all needed tools
 RUN apk update && \
   apk upgrade --update-cache --available && \
-  apk add build-base curl git jq openssh bash
+  apk add build-base curl git jq openssh bash docker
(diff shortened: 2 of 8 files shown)


tests/validation/autoscale_test.go
```diff
@@ -25,7 +25,7 @@ func autoscaleTests(t *testing.T, when spec.G, it spec.S) {
 		it("should autoscale down to 0", func() {
 			// Precondition
 			assert.True(t, service.IsReady())
-			assert.Equal(t, "Hello World", service.GetAppEndpointResponse())
+			assert.Equal(t, "Hi there, I am rio:production6", service.GetAppEndpointResponse())
```
cbron: This is ok ?

rancher-max: Yes... We ignore the no healthy upstream though as part of getting response:
```
	for i := 0; i < 5; i++ {
		if resp == "no healthy upstream" {
			time.Sleep(1 * time.Second)
			resp, _ = GetURL(endpoint)
		} else {
			break
		}
	}
```

cbron:
Besides that it looks like there is some test content still in there, it seems ok. Make sure its up to date with master.

tests/validation/autoscale_test.go
cbron: I meant `rio:production6` seems random

rancher-max: oh we can blame @StrongMonkey for his autoscale image returning that

PR #941, opened 2019-12-17
dont allow buildtag mode and pr mode

cbron:
Issue: https://github.com/rancher/rio/issues/542

Also allow webhook secret to be null as it defaults


cli/cmd/create/build.go
```diff
@@ -14,8 +14,8 @@ func (c *Create) setBuildOrImage(imageName string, spec *riov1.ServiceSpec) erro
 			if c.BuildBranch != "master" {
 				return errors.New("build-branch and build-tag cannot both be set, as build-tag will deploy tags from every branch")
 			}
-			if c.BuildWebhookSecret == "" {
```
cbron: Took this out because people might want to use the default `githubtoken` secret name

Issue #467, opened 2019-09-18
Needs a https redirect feature


ibuildthecloud:
Need to be able to redirect from http to https easily to force specific services/path to https only.

lucidprogrammer:
Say you created a service like this,

`rio run --name sample -p 80:80 nginx:alpine`

```
rio ps
NAME      IMAGE          ENDPOINT                                     PORTS     SCALE     WEIGHT    CREATED          DETAIL
sample    nginx:alpine   https://sample-v0-default.somewhere.on-rio.io   80        1         100%      16 seconds ago  
``` 
Check the virtualservices you got 

```
kubectl get vs
NAME                                 AGE
sample                               110s
sample-tls-somewhere.on-rio.io-tls      110s
sample-v0                            110s
sample-v0-tls-somewhere.on-rio.io-tls   110s

```

Ok, now we need every request coming on the version of your choosing to go to https.

kubectl edit vs sample-v0

Add the following snippet, under the routes.

```

- matcher:
        prefix: /
      redirectAction:
        httpsRedirect: true
      routePlugins: {}
```

Once applied, your spec part of the virtualservice will look something like this


```
spec:
  virtualHost:
    domains:
    - sample-v0-default.somewhere.on-rio.io
    routes:
    - matcher:
        prefix: /
      redirectAction:
        httpsRedirect: true
      routePlugins: {}
    - matcher:
        prefix: /
      routeAction:
        single:
          kube:
            port: 80
            ref:
              name: sample-v0
              namespace: default

```

StrongMonkey:
We are going to add this in router. 

To add a router with HTTPS redirect, run

```
rio router add --https-redirect work to sample_service
```
Then check endpoints of router, access http endpoint will give you redirect link to https

```
$ curl -i http://xxx.on-rio-io          
HTTP/1.1 301 Moved Permanently
location: https://xxx.on-rio-io 
date: Tue, 17 Dec 2019 22:46:04 GMT
server: envoy
content-length: 0
```

To add https redirect to your publicdomain, then register public domain to this router

```
rio route register foo.bar router/work
```

lucidprogrammer:
Thank you @StrongMonkey

rancher-max:
Validated above is working. Small correction to the domain register command: `rio domain register foo.bar router/work`

Issue #936, opened 2019-12-14
Router to publicdomain does not generate ssl cert


rancher-max:
**Describe the bug**
The cert stays as the self signed cert-manager.local when creating a publicdomain to a router, so https is not available.

**To Reproduce**
1. `rio run -n demo -p 80 ibuildthecloud/demo:v1`
2. `rio domain register my.cname.com demo`

**Expected behavior**
Should be able to get https endpoint. It looks like it is available but does not properly create the certificate. 

**Additional context**
`rio -s logs cert-manager`
output:
```
cert-manager-7496c5f7cd-sr7rq cert-manager E1213 21:34:08.266623       1 sync.go:184] cert-manager/controller/challenges "msg"="propagation check failed" "error"="wrong status code '404', expected '200'" "dnsName"="maxrio.qa.rancher.space" "resource_kind"="Challenge" "resource_name"="maxrio.qa.rancher.space-tls-3540917872-3566497347-2877105551" "resource_namespace"="rio-system" "type"="http-01" 
```

rancher-max:
This works now. In testing, noticed that registering a domain to a service app endpoint in one case took over an hour to generate the ssl cert. The cert-manager logs during this time were showing `error"="wrong status code '502', expected '200'"`, but eventually it worked. Also, the cname http endpoint was available during this time, and the https endpoint given by rio, so I believe this is working in v0.7.0-rc2.

PR #942, opened 2019-12-17
Fix router bugs, add HTTPS redirect feature.

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rio/issues/467 https://github.com/rancher/rio/issues/936

docs/router.md
```diff
@@ -35,10 +35,18 @@ Creating a route with path-based matching supports: exact match: `foo`; prefix m
 $ rio route add $name/path to $target
 
 # to add prefix
-$ rio route add $name/path* to $target
+$ rio route add $name/path'*' to $target
```
cbron: Looks like this was just outdated from before ? Didn't realize syntax changed.

modules/gloo/pkg/vsfactory/router.go
```diff
@@ -226,10 +243,6 @@ func addRedirect(route riov1.RouteSpec, gatewayRoute *gatewayv1.Route) {
 		ra.RedirectAction.PathRewriteSpecifier = &gloov1.RedirectAction_PathRedirect{
 			PathRedirect: route.Redirect.Path,
 		}
-	} else if route.Redirect.Prefix != "" {
```
cbron: Prefix is just handled via match now ?

docs/router.md
StrongMonkey: Doesn't work with bash cause you have to escape *.

modules/gloo/pkg/vsfactory/router.go
StrongMonkey: prefix is handled by route.Redirect.Path

PR #944, opened 2019-12-17
dont allow empty service generation

cbron:


cbron:
@StrongMonkey the gitcommit/service change isn't perfect because it still re-creates the pr2 service when pr1 closes, those two should be unrelated events, but at least now it reloads to the right commit.

StrongMonkey:
I think it should be fine, if it loads the right commit then it shouldn't recreate pr2 if pr2 already exists.

PR #945, opened 2019-12-18
Fix port-forward use wrong client loader

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rio/pull/942.

PR #948, opened 2019-12-18
Disable non-template and pr mode together.

cbron:
Issue: #542

### DIFF
--- a/cli/cmd/create/build.go
+++ b/cli/cmd/create/build.go
@@ -2,6 +2,7 @@ package create
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cli/cmd/create/build.go
```diff
@@ -14,6 +15,12 @@ func (c *Create) setBuildOrImage(imageName string, spec *riov1.ServiceSpec) erro
 			if c.BuildBranch != "master" {
 				return errors.New("build-branch and build-tag cannot both be set, as build-tag will deploy tags from every branch")
 			}
+			if c.BuildWebhookSecret == "" {
+				fmt.Println("Warning: tag mode only works with a webhook")
```
cbron: Warning because gitwatcher defaults to watch for a github secret name so we don't want to error out that case, but most users will need this tip.

Issue #607, opened 2019-10-25
Autoscaler not reaching scale 1 if I do HTTP GET and scale was zero


izaac:
**Describe the bug**
The autoscaler seems to be going back to 0 very quick so subsequent GET requests after the first one sometimes fails with `no healthy upstream`

**To Reproduce**
1. Create a service with scale range from zero. `rio run --scale=0-11 -p 80 -n ascaler strongmonkey1992/autoscale:testing`
2. Wait until the endpoint is created and the pods are created and then scale is zero
4. Open the Endpoint in the browser once. Do not close the browser tab. (Chrome 77 was used here).
5. If the content was shown correctly wait until the scale is 0 again. 
6. Try again refreshing the browser once.
7. If no error, repeat 5.

I sometimes get the `no healthy upstream` message.

**Expected behavior**

The endpoint responds with the expected content. Every time I do an HTTP GET with a browser. 

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
v1.14.7-gke.10
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.6.0-alpha.1 (7678d67d)
Rio CLI Version: v0.6.0-alpha.1 (7678d67d)
```
**Additional context**
`rio systemlog`
output:
```
nothing relevant
```

izaac:
This looks as dupe of https://github.com/rancher/rio/issues/584 reported by @rancher-max, please close if this is the case.

davidnuzik:
Also should scale on outgoing requests. (both outbound and inbound)

daxmc99:
This appears to be fixed. I am not seeing any issues and the scale up time has been about 30 sec for the sample image deployed and with images I have tested with. 

Stats for a scale from 0, image pre-pulled. 
```
curl -w "@curl-format.txt" -o /dev/null -s "https://ascaler-v0-default.049910.on-rio.io"
 time_namelookup:  0.032576
       time_connect:  0.032732
    time_appconnect:  0.045256
   time_pretransfer:  0.045323
      time_redirect:  0.000000
 time_starttransfer:  11.383586
                    ----------
         time_total:  11.383646
```

daxmc99:
From a single request point of view from curl this bug is fixed, however there is some finicky behavior in browsers documented in https://github.com/rancher/rio/issues/861. 

This issue has been fixed and there is some more investigation of how browsers perform requests (esp Chrome)

rancher-max:
[...]

PR #952, opened 2019-12-20
Move retry logic for scale from zero into rio

daxmc99:
This moves the retry logic for the scale from zero functionality into
the Rio controller where it can be performed by the Envoy proxy (Gloo).

Addresses https://github.com/rancher/rio/issues/607


stacks/rio-autoscaler-stack.yaml
```diff
@@ -7,7 +7,7 @@ services:
     - '* autoscale.rio.cattle.io/servicescalerecommendations'
     - '* rio.cattle.io/services'
     - '* rio.cattle.io/services/status'
-    image: rancher/rio-autoscaler:v0.1.5
+    image: daxl/rio-autoscaler:dev3
```
daxmc99: Depends on https://github.com/rancher/rio-autoscaler/pull/11

modules/gloo/pkg/vsfactory/vsfactory.go
```diff
@@ -2,6 +2,9 @@ package vsfactory
 
 import (
 	"net"
+	"time"
+
+	"github.com/solo-io/gloo/projects/gloo/pkg/api/v1/options/retries"
```
cbron: nit: space

modules/gloo/pkg/vsfactory/vsfactory.go
```diff
@@ -18,10 +21,16 @@ import (
...
-	rioNamespaceHeader = "X-Rio-Namespace"
+	rioNameHeader         = "X-Rio-ServiceName"
+	rioNamespaceHeader    = "X-Rio-Namespace"
+	scaleFromZeroAttempts = 10
 )
 
+// Retry configuration for the envoy gateway (Gloo)  https://www.envoyproxy.io/learn/automatic-retries
+// for services scaled to zero
+var overallTimeout = time.Second * 30
+var perTryTimeout = time.Second * 2
```
daxmc99: This was what I saw as a decent timeout for each request. Ultimately I think this may need to be configurable based upon the service being scaled

Issue #861, opened 2019-11-22
Investigate intermittent 503's


cbron:
Darren/Dax reported seeing some 503's while clicking around rio services

daxmc99:
`no healthy upstream`  
Reproduction with Chrome: 
- Have a service scaled to zero. 
- Attempt to visit a service endpoint using Chrome with dev tools enabled and disabling the cache
- While waiting for the request to scale, make another request (refresh). 
 - This and subsequent requests will return `no healthy upstream` and 503 errors.

rancher-max:
https://github.com/rancher/rio/issues/584#issuecomment-547027174

daxmc99:
After working with @rancher-max and @StrongMonkey it looks like the root of the problem is that the service version endpoint `$service-v0-on.rio.io` is not being populated correctly. 

The `SSL_SYSCALL` error manifests itself on services scaled to zero, when there is no virtual service for that version endpoint. _Sometimes_, it is dynamically created but the correct behavior should be for the virtual service to *always* exist even if a service is scaled to zero. 

Proposed solution: 
All services in Rio should have a version endpoint regardless of scale.

rancher-max:
This has been fixed in latest master, commit: `c0f480fe`. The scenario this fixes is, as mentioned above:
1. Have a single service on an app, and have it with min scale of 0 and max scale of 1 or greater
2 Wait for the service to be at observed scale of 0, then query the app endpoint (`curl` is fine)
3. While the pod is initializing, make more queries to the app endpoint

This was previously returning `no healthy upstream` on at least one of those subsequent requests, but now returns the correct result:
```
$ for x in {1..20} ; do curl https://other-as-default.7oign7.on-rio.io ; done
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
Hi there, I am rio:production6
```

Issue #584, opened 2019-10-21
Autoscale to 0 does not have healthy upstream after scaling back up


rancher-max:
**Describe the bug**
A service with MinScale set to 0 yields "no healthy upstream" after allowing it to scale down to 0 then scale back up to 1. The first query when it is at 0 observed scale returns the correct response, but the second query (now with the scale at 1) returns the error.

**To Reproduce**
1. Deploy a service with MinScale of 0: `rio run -n noscale -p 80/http --scale 0-4 ibuildthecloud/demo:v1`
2. Query the endpoint to ensure it is initially working: `curl <endpoint>`
3. Wait approx. 3 minutes to allow the service to scale to 0. It is there when its output in `rio ps` shows "0/1" for Scale.
4. Query the endpoint: `curl <endpoint>` -- this should return the expected result of "Hello World"
5. Query again: `curl <endpoint>` -- THIS FAILS. I have seen it succeed a few times, but if you continue to execute `curl` at this point, it will eventually fail. Sometimes it yields "no healthy upstream" and sometimes it gives no response at all.

**Expected behavior**
It should continue to return "Hello World" and should not fail after scaling back up

**Kubernetes version & type (GKE, on-prem)**: `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-19T13:57:45Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.4-k3s.1", GitCommit:"a7531b1ab3fd5ff987b074472ddfb84a2f5326bc", GitTreeState:"clean", BuildDate:"2019-09-19T22:36Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```
Type:
**Rio version**: `rio info`
```
Rio Version: v0.5.0 (303f3652)
Rio CLI Version: v0.5.0 (303f3652)
```

**Additional context**
`rio systemlog`
output:
```
rio-controller | time="2019-10-21T21:22:25Z" level=info msg="Setting desired scale 0 for testing-ns/xbbqf"
rio-controller | time="2019-10-21T21:22:27Z" level=info msg="Setting desired scale 1 for testing-ns/xbbqf"
```

[...]

PR #985, opened 2020-01-04
Fix 503 & 'no healthy upstream' issues

daxmc99:
Address 'no healthy upstream' by ensuring service has allocated corev1.Endpoint IP

https://github.com/rancher/rio/issues/607
https://github.com/rancher/rio/issues/584
https://github.com/rancher/rio/issues/861

### DIFF
--- a/modules/gloo/pkg/vsfactory/target.go
+++ b/modules/gloo/pkg/vsfactory/target.go
@@ -14,9 +14,12 @@ import (
 	"github.com/rancher/rio/pkg/serviceports"
 	"github.com/rancher/rio/pkg/services"
 	"github.com/rancher/wrangler/pkg/name"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 )
 
+// the target struct is used to hold information for networking decisions
 type target struct {
 	Hosts          []string
 	Port           int32
@@ -34,7 +37,8 @@ func (t target) valid() bool {
 	return t.Port != 0 && len(t.Hosts) > 0
 }
 
-func getTarget(obj *riov1.Service, systemNamespace string) (result target, err error) {
+// getTarget returns 1 target for a specified rio service
+func (f *VirtualServiceFactory) getTarget(obj *riov1.Service, systemNamespace string) (result target, err error) {
 	app, version := services.AppAndVersion(obj)
 	result.Name = name.SafeConcatName(app, version)
 	result.Namespace = obj.Namespace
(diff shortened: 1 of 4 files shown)


tests/integration/scale_test.go
```diff
@@ -37,5 +33,33 @@ func scaleTests(t *testing.T, when spec.G, it spec.S) {
...
+		var service testutil.TestService
+		it.Before(func() {
+			service.Create(t, "--scale", "0-3", "sangeetha/mytestcontainer")
+
+		})
+		it("service should never return 503 'no healthy upstream'", func() {
+			// check that service can scale up
+			service.GetEndpointResponse()
+			for service.GetScale() > 0 {
+				time.Sleep(1 * time.Second)
```
daxmc99: @rancher-max Do we have any functions in testutil that can wait till a service naturally scales to zero? Going to add one I think

rancher-max: `service.WaitForScaleDown` should do it for you! You also might want to double check the test in tests/validation/autoscale_test.go that does autoscale down to 0 to make sure you're not duplicating.

I removed it from integration because it was a bit flakey on the "scale back up" part and took more time due to the amount of tests running in parallel

daxmc99:
Dropped the testing commit from this PR.  @rancher-max confirms there is a validation test that covers this. 
Tested locally by curling the endpoint for a service scaled from zero multiple times. Never saw 'no healthy upstream'

modules/gloo/pkg/vsfactory/target.go
```diff
@@ -51,32 +55,32 @@ func getTarget(obj *riov1.Service, systemNamespace string) (result target, err e
...
 		if err != nil {
 			return result, err
 		}
-		if seen[u.Host] {
-			continue
+		if result.ScaleIsZero {
+			if *obj.Status.ComputedReplicas > 0 {
+				logrus.Debug("service has ComputedReplicas > 0 but no IP address allocated via k8s endpoints yet")
+			}
+			result.Port = 80
```
cbron: So basically this whole method first checks if ComputedReplicas is >= 0, and then checks if there are any endpoints. Then if there aren't endpoints it returns a target, thats the actual change here. Previously it only cared about reps. That right ?

modules/gloo/pkg/vsfactory/target.go
```diff
@@ -176,3 +180,57 @@ func formatHost(port, defaultPort int, hostname string) string {
...
+		if len(endpoint.Subsets) != 0 {
+			// ip address allocated for this service, should not be scale zero
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+// determineHosts returns a set of hosts based upon the rio services endpoint URLs
+func determineHosts(hostNameEndpoints []string) ([]string, error) {
```
daxmc99: @cbron Line [65-79](https://github.com/rancher/rio/pull/985/files#diff-0586357b98c09d1ed2a1132e44706c4dL65-L79) was moved into this function

modules/gloo/controller/service/handler.go
```diff
@@ -23,6 +23,8 @@ func Register(ctx context.Context, rContext *types.Context) error {
 		rContext.Rio.Rio().V1().Service(),
 		rContext.Apply.WithCacheTypes(
 			rContext.Gateway.Gateway().V1().VirtualService(),
+			// TODO decide if I need this
+			rContext.Core.Core().V1().Endpoints(),
```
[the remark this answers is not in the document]
daxmc99: It works fine without it... I thought it may be needed since there is a requirement on looking up `corev1.Endpoints` but I do it with a direct API call. @StrongMonkey may need to look at this

modules/gloo/pkg/vsfactory/target.go
cbron: Now we wait for IPs _and_ computedReplicas of >= 0. Not nil computedReplicas because we want it to have been scaled at least once and then downscaled in order to get the IP ? Is there a reason we need both checks ? 

`isScaleZero` feels unintuitive since its not really dealing with scale specifically, its dealing with the "zero-scale" state which affects other fields/resources. Wish we had a separate word for that. Could rename it to `noEndpoints`, or `noIP`, or just `isZeroScaled`, but I guess its fine as-is also.

daxmc99: nil `computedReplicas` is an error state since nil `computedReplicas` indicates that the service is not a canidate for autoscaling. If you have nil `computedReplicas` and a scale of zero there is an issue outside of this router code.

daxmc99: > Now we wait for IPs and computedReplicas of >= 0
If computeReplics >= 0 we only print a debug log. No impact on logic

modules/gloo/pkg/vsfactory/target.go
```diff
@@ -51,32 +55,32 @@ func getTarget(obj *riov1.Service, systemNamespace string) (result target, err e
...
-		result.Port = 80
-		result.OriginalTarget = struct {
-			Name      string
-			Namespace string
-		}{Name: obj.Name, Namespace: result.Namespace}
-		result.Name = constants.AutoscalerServiceName
-		result.Namespace = systemNamespace
+	result.Hosts, err = determineHosts(obj.Status.Endpoints)
+	if err != nil {
+		logrus.Warnf("%v", err)
```
StrongMonkey: Is it for debugging purpose? I think the error will ultimately pop up and show up in logs.

daxmc99: Dropped

Issue #993, opened 2020-01-14
Staging service fails with namespace in name rather than as param


cbron:
Only occurs in non default namespace

```
rio run -n x:demo nginx
rio stage x:demo v2
 -> x:demo@v2
 -> FATA[0000] failed to find service demo
```

rancher-max:
Working in latest master, commit: c0f480fe. Validated issue scenario and did a regression on default namespace, promoting, arbitrary weights, and a few other basic stage commands with setting image.

PR #994, opened 2020-01-14
cli utils bug with namespace-in-name

cbron:
Issue: https://github.com/rancher/rio/issues/993

The `ListAppServicesFromServiceName` is incorrect because it won't take into account namespace-in-name, and redundant because `ListAppServicesFromAppName` gives you same data.

### DIFF
--- a/cli/cmd/util/service.go
+++ b/cli/cmd/util/service.go
@@ -7,19 +7,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func ListAppServicesFromServiceName(ctx *clicontext.CLIContext, serviceName string) ([]*riov1.Service, error) {
-	namespace := ctx.GetSetNamespace()
-	service, err := ctx.ByID(serviceName)
-	if err != nil {
-		return nil, err
-	}
-
-	svc := service.Object.(*riov1.Service)
-	app, _ := services.AppAndVersion(svc)
-	return ListAppServicesFromAppName(ctx, namespace, app)
-
-}
-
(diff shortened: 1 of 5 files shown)


cli/cmd/util/service.go
```diff
@@ -7,19 +7,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func ListAppServicesFromServiceName(ctx *clicontext.CLIContext, serviceName string) ([]*riov1.Service, error) {
```
daxmc99: Nice 😎

cli/cmd/util/service.go
```diff
@@ -7,19 +7,6 @@ import (
...
-	if err != nil {
-		return nil, err
-	}
-
-	svc := service.Object.(*riov1.Service)
-	app, _ := services.AppAndVersion(svc)
-	return ListAppServicesFromAppName(ctx, namespace, app)
-
-}
-
```
daxmc99: Nice 😎

tests/testutil/testutil.go
```diff
@@ -61,6 +61,11 @@ func ValidationPreCheck() {
 // Example: args=["run", "-n", "test", "nginx"] would run: "rio --namespace testing-namespace run -n test nginx"
 func RioCmd(args []string, envs ...string) (string, error) {
 	args = append([]string{"--namespace", TestingNamespace}, args...)
+	return RioExec(args, envs...)
+}
+
+// RioExec executes rio CLI commands with arguments, use RioCmd unless you have to use non testing ns
+func RioExec(args []string, envs ...string) (string, error) {
```
rancher-max: Can we name this function something different? It's confusing since there is a `rio exec` command as well

cbron: Yes, fixed.

PR #1006, opened 2020-01-29
Edit help/usage to be more consistent and contain examples

marcinkoziej:
This patch contains edits / fixes to rio help / usage system.
Beware, I am not a go programmer; I tried not to break the syntax but I do not know it nor style best practices.

Why's:
- everywhere where I encountered a $FOO or $Foo placeholder, I replaced it with FOO
- build-history - use phrase "build history" with a space in usage.
- config - specify that configmap/ must be prefixed to name. Using NAMESPACE/NAME does not work. One needs to do
- route - removed the `-n NAMESPACE` because this applies to all commands
- secret - the subcommand `ls` does not accept options nor arguments, do not state otherwise
- exec - changed CONTAINER to SERVICE because the argument is what `rio ps` displays, what seems to be called a SERVICE. The container can be specified by `-c` option
- uninstall - does not accept options

Last but not least, **thank you for amazing work** :heart:. I've been a fan since Rancher 1!

### DIFF
--- a/cli/cmd/buildhistory/buildhistory.go
+++ b/cli/cmd/buildhistory/buildhistory.go
@@ -18,7 +18,7 @@ func History(app *cli.App) cli.Command {
 		"")
 	return cli.Command{
 		Name:   "build-history",
-		Usage:  "Show build-history",
+		Usage:  "Show build history",
 		Action: clicontext.DefaultAction(hist.Action),
 		Flags:  table.WriterFlags(),
 	}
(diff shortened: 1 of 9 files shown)


StrongMonkey:
@marcinkoziej Thanks for the contribution! Changes looks good, but looks like you have one file that haven't been formatted thus CI failed.

marcinkoziej:
@StrongMonkey fixed the formatting. However, some other error appeared. Seems not related

StrongMonkey:
@marcinkoziej Thank you, looks like a flaky test.

rancher-max:
Yep, that test is one of the flakes I'm currently working on fixing. Changes LGTM!

PR #1012, opened 2020-02-05
Add bash/zsh completion

daxmc99:
This PR enables shell completion via urfav
In order to use it, you must source the correct shell completion script
in cli/autocomplete/ with the env variable PROG=rio.
For example, 
`PROG=rio source $RIO_REPO_ROOT/cli/autocomplete/bash_autocomplete`

There is still more to be done in order to make this consumable by end
users.


daxmc99:
@StrongMonkey PTAL 😄

StrongMonkey:
@daxmc99 Will do :+1:

Issue #954, opened 2019-12-22
no option to specify secret name


oskapt:
Note: Please add dashboard specific bugs to the dashboard repo: https://github.com/rancher/dashboard/issues

**Describe the bug**

The rio commands create a secret of a predefined name and assume that running a workload will always use the same secret. See [this page](https://github.com/rancher/rio/blob/master/docs/continuous-deployment.md#private-github-repo) where it states:

> Rio will automatically use this secret now when doing standard git checkout, so no additional flags need to be passed. 

While I _can_ specify which secret to use when using `rio run`, there is no non-manual way to create a secret with a different name.

A side-effect of this is also that `rio secret` clobbers the same secret when run multiple times.

**Expected Behavior**

I would like to be able to create multiple secrets for different private repos (gitlab/github/personal), update them, and specify which secret to use when building/deploying applications within Rio. This could be solved by permitting me to name the secret when creating/updating it.

I also think that Rio shouldn't assume which secret I want to use. If I'm using a per-repo deploy token or deploy key, does it know to use the key for `gitlab.com/monachus/private-project` and not to use it with `gitlab.com/monachus/other-project` where `other-project` might need a different key? 

If no key is provided in the command, it should not use a key and should fail because it did not have permission to access the repo. If a key is provided in the command, it should use the key. This is more in line with the command-line instructions provided by the user.

PR #1013, opened 2020-02-06
Add option to specify secret name

rancher-max:
#954 

Also update bug template to use comments to that submitted bugs are cleaner

### DIFF
--- a/cli/cmd/secrets/create.go
+++ b/cli/cmd/secrets/create.go
@@ -16,6 +16,7 @@ import (
 )
 
 type Create struct {
+	N_Name        string   `desc:"Assign a name to the secret. Use format [namespace:]name"`
 	T_Type        string   `desc:"Create type" default:"Opaque"`
 	F_FromFile    []string `desc:"Creating secrets from files"`
 	D_Data        []string `desc:"Creating secrets from key-pair data"`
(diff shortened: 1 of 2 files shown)


cli/cmd/secrets/create.go
```diff
@@ -30,6 +31,13 @@ const (
 )
 
 func (s *Create) Run(ctx *clicontext.CLIContext) error {
+	name := s.N_Name
+	if name != "" {
+		r := ctx.ParseID(fmt.Sprintf("secret/%s", name))
+		if _, err := ctx.Core.Secrets(r.Namespace).Get(r.Name, metav1.GetOptions{}); err == nil {
+			fmt.Printf("Warning: %s already exists\n", r)
```
cbron: Shouldn't this be `err != nil` ? Also I would check against AlreadyExists error type, this could be a diff error.

rancher-max: `err == nil` in this case because it returns no error if a secret with this name exists.

cbron: Ahh, right.

cbron: You're printing the v1.Secret type ? Why not just name.

rancher-max: I was just printing similar to how we do it elsewhere. It ends up being the name with the type prepended:
```
$ ./bin/rio secret create -n githubtoken --docker
Warning: secret/githubtoken already exists
```

StrongMonkey:
@rancher-max actually....Can we make the name also a prompt similar to other options?

rancher-max:
@StrongMonkey added prompt as an additional option. It will prompt if the `-name` flag is not specified, otherwise it will use the value declared in the command. It only gives a warning if using the `-name` flag with a name that already exists, otherwise it will overwrite as in current state.

Issue #1031, opened 2020-04-23
requestTimeoutSeconds only affects version specific services in Gloo


r2ewj:
I created a new service via Riofile with a timeout:

services:
  ido-ui:
    requestTimeoutSeconds: 120

This creates 4 VirtualServices in Gloo, I've noticed that the version specific services receive the Timeout option, but the main end point does not.
ido-ui-v0 has the timeout
ido-ui-v0-tls-#####.on-rio.io-tls has the timeout
ido-ui-tls-####.on-rio.io-tls no timeout
ido-ui no timeout
This is the expected result from ido-ui-v0:
spec:
  virtualHost:
    domains:
    - ido-ui-v0-rf.####.on-rio.io
    routes:
    - matchers:
      - prefix: /
      options:
        timeout: 120s
      routeAction:
        single:
          kube:
            port: 80
            ref:
              name: ido-ui-v0
              namespace: rf

This is the result from the non versioned service
spec:
  virtualHost:
    domains:
    - ido-ui-rf.#####.on-rio.io
    routes:
    - matchers:
      - prefix: /
      routeAction:
        single:
          kube:
            port: 80
            ref:
              name: ido-ui-v0

StrongMonkey:
Should be fixed in v0.7.1-rc2.

StrongMonkey:
@r2ewj I will close this issue since it should have been fixed in v0.7.1. Feel free to reopen if you still see it.

PR #1035, opened 2020-05-04
Set requestTimeout for app virtualservice

StrongMonkey:


StrongMonkey:
Fix https://github.com/rancher/rio/issues/1031.

